PR #66, opened 2026-05-03
Allow R2 presigned upload endpoint in CSP

Somu878:
## Summary

Fixes web pet uploads failing with:

`Upload failed: R2 PUT network error: xhr network error`

The upload flow uses presigned R2 PUT URLs on `*.r2.cloudflarestorage.com`, but the app’s Content Security Policy only allowed the public R2 read host. Because of that, the browser could block the direct PUT request before R2 returned a normal response.

## Fix

Added `https://*.r2.cloudflarestorage.com` to `connect-src` in the CSP so browser-based presigned R2 uploads are allowed.

## Testing

- Ran `biome check next.config.ts`
- Confirmed the R2 signing endpoint uses `r2.cloudflarestorage.com`

### DIFF
--- a/next.config.ts
+++ b/next.config.ts
@@ -11,8 +11,8 @@ import type { NextConfig } from "next";
 // - clerk.petdex.crafter.run + *.clerk.com / *.clerk.accounts.dev for
 //   the Clerk client SDK
 // - vercel-scripts / vitals for Vercel analytics
-// - R2 public bucket + UploadThing host + Clerk image hosts + social
-//   avatar hosts for sprites and avatars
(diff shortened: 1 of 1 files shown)


Railly:
Same fix landed independently in 141d58e a few minutes ago. Thanks for the catch and the clean diagnosis — the R2 PUT endpoint missing from connect-src was indeed the root cause of the wave of upload failures. Closing this in favor of the merged hotfix.

If you want a Petdex pet with your credit, send one in once the fix has rolled out (Vercel is finishing the deploy). Real thanks for the fast write-up.

PR #97, opened 2026-05-05
fix: make contributor tests server-safe

Ethan2298:
## Summary

This PR makes the default contributor test suite pass without requiring a Next server graph or a live database, while keeping server-only production modules protected.

- extracts pure install script rendering and submission validation helpers for unit tests
- keeps `install-script.ts` and `submissions.ts` as server-only modules with top-level DB imports
- moves DB-backed search coverage to an explicit `test:db` integration command
- makes the POSIX shell syntax test use `sh` from PATH instead of hardcoding `/bin/sh`
- documents the default and DB-backed test commands for contributors

## Why

`bun test` currently imports DB/server-only code through test files, which can fail before the useful regression tests run. The DB search tests also need a real `DATABASE_URL`, so keeping them in default discovery makes the local contributor path brittle.

## Validation

- `bun test` passes: 41 pass, 0 fail
- `bun --bun tsc --noEmit` passes
- touched-file Biome check passes
- `bun run test:db` fails explicitly without `DATABASE_URL`, as intended

`bun run build` compiles and completes TypeScript, then fails during page data collection because this local shell has no Neon `DATABASE_URL` configured.

`bun run check` still reports existing repo-wide CRLF/formatting and lint diagnostics outside this patch.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -55,7 +55,8 @@ This path is for maintainers and trusted collaborators.
 
 - **Runtime**: Bun (never `npm install`).
 - **Lint/format**: `bun run check` / `bun run format` (Biome).
-- **Tests**: `bun test`.
+- **Tests**: `bun test`. DB-backed search integration tests are explicit:
+  `DATABASE_URL=... bun run test:db`.
 - **Commit style**: conventional commits (`feat:`, `fix:`, `docs:`, etc.).
 
 ## Where to look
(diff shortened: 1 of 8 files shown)


Railly:
Merged. Splitting the pure helpers into install-script-render.ts and submissions-validation.ts, plus the server-only directives, keeps the default `bun test` runnable for contributors without a DATABASE_URL. The explicit `test:db` command is the right escape hatch. Verified locally: 41 pass, 0 fail, tsc clean.

PR #101, opened 2026-05-05
fix(my-pets): prevent action menu clipping

iambinlin:
## Summary

- Prevent the My Pets action menu from being clipped by submission card bounds.
- Keep clipping constrained to the pet sprite stage so animated sprites still stay within their visual area.
- Raise the active card while menu controls have focus so the opened menu appears above neighboring cards.

## Root Cause

The submission card used `overflow-hidden` on the outer card container. The detail action menu is absolutely positioned inside that card, so any portion extending below the card edge was clipped.

## Validation

- `./node_modules/.bin/biome check src/components/my-pets-view.tsx`
- `git diff --check`

Note: the full `npm run check` was not used as a gate for this PR because it reports pre-existing lint/format issues outside this change scope.

### DIFF
--- a/src/components/my-pets-view.tsx
+++ b/src/components/my-pets-view.tsx
@@ -297,7 +297,7 @@ function SubmissionCard({ submission }: { submission: Submission }) {
 
   return (
     <article
-      className={`group relative flex flex-col overflow-hidden rounded-3xl border bg-surface/76 backdrop-blur transition ${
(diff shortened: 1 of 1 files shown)


Railly:
Merged. Thanks ryanwang. The overflow-hidden on the outer card was clipping the absolute-positioned menu, and your two-line fix moves the clipping down to the sprite stage where it actually belongs.

PR #108, opened 2026-05-06
Add automated submission review autopilot

cuevaio:
## Summary
- Add live automated submission review for web and CLI submissions, including asset validation, visual/hash duplicate detection, semantic duplicate checks, policy holds, and auto-approve/auto-reject decisions.
- Add review storage/migrations, Gemini embedding backfill/review retry scripts, admin evidence UI, status filters, rerun action, and localized submit outcomes.
- Harden rollout after review: preserve hard takedown, add review-performance indexes, bound dHash scanning, make review backfills idempotent by default, raise initial auto-approve confidence to 0.95, add submit latency analytics, and update CLI profile links to `/u/<handle>`.

## Migration Notes
- `0005_gemini_embeddings.sql` intentionally clears old 1536-dim text embeddings when switching to the required 3072-dim Gemini image-aware embedding model.
- Any environment that already ran `0004` with populated embeddings should rerun `bun scripts/compute-similarity.ts --force` after applying `0005`/`0007`.
- `0007_submission_review_indexes.sql` is included in `drizzle/meta/_journal.json`; `npx drizzle-kit migrate` now sees it and was verified locally.

## Verification
- `npx drizzle-kit migrate`
- `bunx biome check` on modified root files and `packages/petdex-cli/bin/petdex.ts`
- `bun test --env-file=.env.mock src/lib/embeddings.test.ts src/lib/submission-review.test.ts src/lib/security.test.ts`
- `bun test --env-file=.env.mock`
- `bun run i18n:check`
- `bun --env-file=.env.mock run build` (passes with existing Turbopack NFT warnings)
- `packages/petdex-cli`: `bun run build && bun run typecheck`


Railly (changes requested):
Did a deep local review. Strong shape overall, the auto-review pipeline is well-decomposed (policy + duplicates + decisions split across small files), the migrations are sane, and the test surface for `embeddings.ts` and the decision logic is solid. Couple of blockers and a few things worth fixing before this lands.

## Blockers

**1. The takedown DELETE handler in `src/app/api/admin/[id]/route.ts` is gone.**

This PR was branched before commit 9be8559 which added `DELETE /api/admin/[id]` (hard takedown: removes the pet row, every cross-table reference, R2 assets, sends owner email). On this branch the file only exports `PATCH`. After rebase that handler needs to be re-added by hand — the refactor that extracts `applySubmissionAction` didn't carry the DELETE path along.

Concrete check on this branch: `grep "export async function" src/app/api/admin/[id]/route.ts` returns one match. On main it returns two.

**2. `findVisualMatches` does a full table scan on every submission review.**

`src/lib/submission-review.ts` lines 579–618 fetches every approved+pending row with `dhash IS NOT NULL` into Node, computes hamming distance in JS. No DB-side filter, no index on `dhash`. Fine at current scale. At a few thousand submissions this becomes a multi-second scan that can blow past the 30s `SUBMISSION_REVIEW_TIMEOUT_MS` and cause every review to fall through to `hold`.

Pragmatic fix: chunked SQL query + early exit when a match below threshold is found. Or push hamming into SQL with a custom function. At minimum add a partial index on `dhash WHERE status IN ('approved','pending')` so the scan is cheap.

## Major

**3. Migration 0005 destructively wipes 1536-dim embeddings.**

`drizzle/0005_gemini_embeddings.sql` has `ALTER COLUMN embedding TYPE vector(3072) USING NULL::vector(3072)`. Expected per the comment, but it's a silent data loss if 0004 has ever run against staging populated data. Worth a runbook note on PR description: "if you've already applied 0004 in any environment, rerun the embedding backfill after 0005 to repopulate."

**4. Embedding call is in the hot path of `/api/submit`.**

Every submit waits for a Gemini AI Gateway round-trip plus a `gpt-5-mini` policy call before returning 201. `SUBMISSION_REVIEW_TIMEOUT_MS = 30s` caps the worst case but every successful submit eats the full latency. Consider making review async (insert pet as `pending` + `submission_reviews` row as `running`, run review off-thread, finalize via the internal endpoint). The 201 response can fire instantly and the admin queue picks up the result later. Simpler design: same response shape, but predictable latency.

**5. Backfill script is not fully idempotent.**

`scripts/review-pending-submissions.ts` calls `reviewSubmission(row.id, { force: true })` unconditionally. If the script crashes mid-batch and is rerun, submissions already reviewed get a second review record. The `force` option exists but isn't actually consumed by `reviewSubmission` (`_options` is unused). Either wire it through, or have the script skip submissions that already have a `submission_reviews` row with `status='completed'`.

## Nits

- `REVIEW_MODEL = "openai/gpt-5-mini"` and `PETDEX_EMBEDDING_MODEL` are both hardcoded strings. Make them env-overridable so we can A/B without redeploying.
- `finishReview` throws if the update returns no rows. Can't happen in normal flow, but a connection pool blip between `insertReview` and `finishReview` produces a stranded review record with `status='running'` forever. Worth a janitor query in the backfill to bury those.
- `cmdSubmit` in `packages/petdex-cli/bin/petdex.ts` prints `"Track approval at ${PETDEX_URL}/my-pets"`. `/my-pets` was deleted in the recent profile refactor on main; CLI users will get a 404. Update to `/u/<their-handle>` (or `/`) after rebase.
- `auto-approve` requires `policy.decision === 'pass'` and `confidence >= 0.9`. Sound choice, but for the first week of live mode I'd ratchet it to 0.95 and watch the manual override rate before relaxing.

## What I'd want before merge

1. Re-add `DELETE /api/admin/[id]` handler (blocker)
2. Fix the dhash index or chunked scan (blocker)
3. Decide async vs sync review path (can ship sync now if you accept the latency cliff and add a Vercel Analytics event for time-to-201)
4. Wire `force` through or fix backfill idempotency
5. Bump auto-approve confidence to 0.95 for the first week
6. Update CLI submission tracking URL to `/u/<handle>` after rebase

Reviewer agent's full notes saved locally if you want the longer version. Solid PR overall, just needs the takedown handler back and the dhash scan tightened.

cuevaio:
Thanks for the deep review. Addressed the actionable items in `84f7be8`.

What changed:
- Hard takedown: this was stale after the latest merge. `DELETE /api/admin/[id]` is present again alongside `PATCH`, including slug-reference cleanup, R2 cleanup, and owner notification.
- dHash scan: added `0007_submission_review_indexes.sql` plus the missing `drizzle/meta/_journal.json` entry, so Drizzle now applies it. `findVisualMatches` now scans in bounded chunks and fails closed to hold/manual review if the candidate cap is reached.
- Backfill idempotency: `reviewSubmission(..., { force })` now consumes `force`; default runs reuse the latest review instead of creating duplicates. `scripts/review-pending-submissions.ts` now only forces with `--force` and buries stale `running` reviews.
- Auto-approve threshold: bumped from `0.9` to `0.95` and added a regression test for `0.94` holding.
- CLI URL: replaced `/my-pets` output with the canonical `/u/<handle>` profile URL from the submit response, with a safe fallback.
- Submit latency: kept sync review for this PR but added upload/register/total duration analytics so we can measure the actual cliff before introducing queue/polling architecture.
- Embeddings: keeping the 3072-dim Gemini image-aware embedding migration intentionally. Added a runbook note to rerun `bun scripts/compute-similarity.ts --force` after `0005` in any env where `0004` had populated old 1536-dim embeddings.

Intentionally deferred:
- Full async review path. The current UI/CLI relies on immediate outcome and the bounded sync path is now instrumented; we can move to background review once latency data says it is worth the extra workflow complexity.
- Env-overridable review/embedding models. We are keeping project constants for this rollout.

Verified:
- `npx drizzle-kit migrate`
- `bunx biome check` on modified files
- `bun test --env-file=.env.mock src/lib/embeddings.test.ts src/lib/submission-review.test.ts src/lib/security.test.ts`
- `bun test --env-file=.env.mock`
- `bun run i18n:check`
- `bun --env-file=.env.mock run build`
- `packages/petdex-cli`: `bun run build && bun run typecheck`

Railly (approved):
Round 2 review. All five blockers/majors from the first round are resolved cleanly. Approving with two small notes.

## Closed

**1. DELETE handler restored** — `grep "export async function" src/app/api/admin/[id]/route.ts` returns both `PATCH` and `DELETE` again. ✓

**2. `findVisualMatches` chunked** — pages in groups of 250, caps at 2000 rows, returns a `VisualMatchScan` with a `complete` flag plus a truncation reason. The new index in `drizzle/0007_submission_review_indexes.sql` covers `(status, created_at DESC) WHERE dhash IS NOT NULL` which makes each chunk fast. Full Node-side table scan is gone. ✓

**3. Migration 0005 documented** — the destructive `vector(1536) -> vector(3072)` is intentional for the Gemini move. Comment on `scripts/compute-similarity.ts` tells operators to re-run with `--force` afterwards. Make sure that line is in the deploy runbook so the catalog isn't left without semantic embeddings until the next nightly. ✓

**4. Backfill idempotent by default** — `reviewSubmission` early-exits with `{ reused: true }` when a review exists and `force` is false. The script's `--force` flag is now actually wired through. Re-running after a crash is safe. ✓

**5. CLI no longer points at `/my-pets`** — both the "nothing new" early-exit path and the final summary outro now use the `profileUrl` from the API response (or a pre-flight `whoami()` fallback). ✓

## Notes for the deploy

- **0007 indexes are not CONCURRENTLY**. Drizzle migrations run inside a transaction so `CONCURRENTLY` isn't allowed there. On a populated production table this means three exclusive index locks on `submitted_pets`. Either run these during off-peak or pre-run them manually with `CREATE INDEX CONCURRENTLY` outside the transaction and then apply the migration as a no-op via Drizzle's hash-tracking. Coordinate with the deploy sequence.

- **`compute-similarity.ts --force` after 0005**. Easy to forget. Add to the runbook so the catalog isn't left empty of semantic embeddings between deploy and the next nightly.

## Major still open (your call to ship anyway)

The review path stays synchronous on `/api/submit` with a 30s timeout — fast submissions complete in well under that, but at the cap users see a 30s spinner before getting a response. You've added `register_duration_bucket` analytics so this is observable; if the data shows it sitting comfortably under 5s in practice, ship as-is and do an async refactor only if real submitters start hitting the ceiling. A non-blocker for me.

## Nits (skip if you want)

- `findVisualMatches` does up to 8 sequential round-trips. With the new index each is cheap so this is fine at current scale; if submissions grow past ~10k consider replacing with a stored Hamming function or a pre-computed sorted bigint column for one-shot SQL filtering.
- `userProfileUrl` in the CLI falls back to `user.sub.slice(-8)` when neither `preferred_username` nor `username` is present. Only affects the "nothing new" early-exit path since the API response provides a real profileUrl on successful submit. Worth a comment but not a fix.

Solid PR. Approving — squash whenever you're ready.

Railly:
Pushed `b00cfaa` directly to this branch — adds a `political_public_figure` category to the policy so the auto-review holds parodies of current heads of state / sitting politicians / royalty / contemporary news figures for human review.

Context: ryanwang opened feedback `fb_fb600918b04747aea9` flagging 7 pets (trump/trumpet/ikun/elon/xi-jinping/king-charles/princess-kako). Took those down manually. This locks the rule into the prompt so the next batch gets caught before reaching the queue.

Threshold 0.4 — strict enough to actually catch celebrity/leader sprites, lenient enough that genuine fictional characters with political vibes don't get held by accident. Historical figures (Confucius, Laozi, etc.) explicitly carved out in the description.

Test suite still 8/8 pass. Ready when you want to merge.

PR #167, opened 2026-05-08
Fix clipped locale switcher menu

henryjing96:
## Summary

- Render the locale switcher options through the existing `Popover` portal so the menu is no longer clipped by the parent user dropdown.
- Keep the existing language switcher UI while replacing the local absolutely-positioned menu.
- Add a `document.cookie` fallback for `NEXT_LOCALE` when `cookieStore` is unavailable.

## Validation

- `bun run check`
- `bun run i18n:check`
- Manual QA with `dev:mock`: verified the user-menu language options render fully and switching `/zh -> / -> /es` works.

### DIFF
--- a/src/components/locale-switcher.tsx
+++ b/src/components/locale-switcher.tsx
@@ -1,13 +1,19 @@
 "use client";
 
 import { usePathname, useRouter, useSearchParams } from "next/navigation";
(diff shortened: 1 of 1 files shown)


cuevaio:
Thank you, @henryjing96!

cuevaio (approved):
Thank you, @henryjing96!

PR #215, opened 2026-05-10
fix: render surprise pet thumbnails correctly in mock mode

henryjing96:
## Summary
- Resize and clip the Surprise Drop pet thumbnail so oversized sprite frames stay inside the card.
- Improve `dev:mock` seeded pet assets by using real curated R2 spritesheets when available.
- Generate sprite-grid-compatible mock placeholders for pets without real assets, avoiding blank or black thumbnails during local QA.

## Validation
- `bun x biome check src/lib/mock/db.ts src/components/surprise-pet-card.tsx`
- `bun run i18n:check`
- Manual QA with `bun run dev:mock` on `localhost:3001`
- Verified Surprise Drop thumbnails render images and stay inside the card.
- Verified mock fallback sprites render for pets without real R2 spritesheets.

## Notes
- The mock placeholder change only affects `dev:mock`; production, real DB data, and real R2 assets are unchanged.
- `bun run check` currently fails on upstream Biome issues in unrelated CLI/desktop/drizzle files

## Before and after
surprise-drop-bug
surprise-drop-fix

### DIFF
--- a/src/components/surprise-pet-card.tsx
+++ b/src/components/surprise-pet-card.tsx
@@ -23,6 +23,7 @@ type SurprisePet = {
 };
 
 const BASE_KEY = "petdex_surprise_pet_seen";
+const THUMBNAIL_SPRITE_SCALE = 0.45;
 
 export function SurprisePetCard() {
   const locale = useLocale();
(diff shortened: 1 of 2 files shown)


henryjing96:
Hi! Just wanted to check in on this PR. Is there anything I should add or adjust before it can be reviewed/merged?

Railly:
Hey @henryjing96, thanks for the fix. The thumbnail clip is exactly right and the mock asset upgrade is great for local QA.

Two things I noticed before merging that I'm going to push directly to your branch (since `maintainerCanModify` is on, credit stays yours):

1. **Accidental revert**. Your branch is 76 commits behind main and the diff drops the `safeGetItem` / `safeSetItem` calls in `surprise-pet-card.tsx`. Those helpers were added in #164 to keep the card from crashing in Safari private browsing and locked-down mobile contexts. Going to rebase + restore them.
2. **Mock url shape**. For non-curated mock pets the new code sets `pet_json_url` and `zip_url` to the SVG fallback data URI. Works for thumbnails but `dev:mock` users testing install/download would get SVG markup instead of JSON/zip. Going to set those two to `null` for the non-curated branch so the failure mode is honest.

Will push in a few minutes.

Railly:
Pushed `5b8c319`. Diff is now:

1. `977ddb0` (your commit, rebased clean on main): thumbnail clip + scale + mock seed assets — unchanged from your work
2. `5b8c319` (mine, on top): switch non-curated `pet_json_url` / `zip_url` to `mock://no-asset/...` sentinel since the schema is NOT NULL

Both `bun run build` and `bun run i18n:check` pass locally. The Vercel preview check shows FAILURE because forks need manual auth from a Crafter Station member (the bot mentioned this on your first push); will trigger that and merge once opencode-review finishes its rerun.

Issue #178, opened 2026-05-09
petdex CLI login fails with Clerk invalid_client


Hakurei-git:
## Summary

The Petdex CLI OAuth login currently fails before account sign-in. Clerk rejects the authorization request with `invalid_client`, so users cannot authenticate and therefore cannot submit pets.

## Command

```powershell
npx.cmd -y petdex@latest login
```

I also tried clearing the npm cache and running the latest CLI again:

```powershell
npm.cmd cache clean --force
npx.cmd -y petdex@latest login
```

## Actual result

The browser opens the Clerk authorization URL, but the page immediately returns:

```json
{
  "error": "invalid_client",
  "error_description": "Client authentication failed (e.g., unknown client, no client authentication included, or unsupported client authentication method). The requested OAuth 2.0 Client does not exist."
}
```

## Expected result

The browser should show the Clerk sign-in/authorization flow, then redirect back to the local CLI callback URL so `petdex whoami` works and `petdex submit <pet-folder>` can upload a pet.

## Environment

- OS: Windows
- - Node: v24.14.0
- - - CLI: `petdex@latest` (previous help output showed `petdex 0.1.2`)
- - - - Shell command used: `npx.cmd` because PowerShell blocks the `.ps1` shim on this machine
## Notes

This appears to happen before any user login step, so it looks like the OAuth client ID used by the published CLI is not configured or no longer exists in Clerk.

Railly:
Hey @Hakurei-git, sorry for the delay. Found the root cause and fixed it on our end.

**What was wrong**: the OAuth client in Clerk had two config issues. The 'Public' toggle (which lets PKCE-only clients exchange codes without a client_secret) was off, and the registered redirect URI was literally `http://127.0.0.1:0/callback` instead of a wildcard pattern. The CLI uses a kernel-assigned ephemeral port like `127.0.0.1:54321/callback`, so the string match never succeeded — Clerk returned `invalid_client`.

**What we did**:
- Enabled Public on the OAuth app (PKCE flow now allowed without a client_secret).
- Cleaned up the redirect URI to `http://127.0.0.1/callback`. Clerk treats loopback URIs at `127.0.0.1` as wildcard-port, so any ephemeral port the CLI grabs will match.

**Smoke test**: I just ran `petdex login` end-to-end on a fresh logout. Got a random port (`63667` this time), the browser flow completed, tokens landed in the keychain, `petdex whoami` works.

Could you retry on your side?

```powershell
[...]

PR #232, opened 2026-05-11
ci(cli): add cross-platform and OAuth smoke checks

iambinlin:
## Summary

This PR adds CI coverage for the Petdex CLI failure mode reported in #178.

It adds two checks:

- `cli-ci`: runs the CLI package across Ubuntu, Windows, and macOS with Node 20, 22, and 24.
- `cli-auth-smoke`: runs a scheduled/manual production OAuth authorize smoke check that fails if Clerk rejects the public CLI OAuth client with `invalid_client` before sign-in.

## Why

Issue #178 reports `petdex@latest login` failing on Windows before the user can sign in because Clerk returns `invalid_client`.

A normal cross-platform build matrix is useful for catching Windows shell, Node version, and packaging issues, but it cannot fully catch this particular failure by itself because `invalid_client` is caused by the public OAuth client configuration used by the CLI.

This PR therefore adds both layers:

- package-level CI for the CLI runtime surface
- production OAuth configuration smoke coverage for the external Clerk authorize path

## What Changed

- Added `.github/workflows/cli-ci.yml`
  - installs CLI dependencies with Bun
  - builds the CLI
  - runs typecheck
  - runs `bun test` on Unix runners where the current test suite is portable
  - skips the full test suite on Windows because several existing CLI tests assert POSIX-only process, shell, path, and chmod behavior
  - verifies the built binary with `node dist/petdex.js --version`
  - packs the npm tarball
  - verifies the packed CLI through `npx` on Unix runners and `npx.cmd` on Windows runners

- Added `.github/workflows/cli-auth-smoke.yml`
  - runs every 6 hours, on manual dispatch, and after relevant pushes to `main`
  - uses Node 24
  - calls the public production CLI auth config endpoint
  - starts a loopback callback server on `127.0.0.1:0`, matching the real CLI's dynamically assigned callback port
  - builds a PKCE authorize URL matching the CLI login flow
  - fails if the authorize response contains `invalid_client`

- Added `packages/petdex-cli/scripts/oauth-authorize-smoke.mjs`
  - keeps the OAuth smoke logic out of inline YAML
  - stops before any user sign-in
  - does not use secrets
  - does not print the OAuth client ID
  - closes the local callback server after the authorize smoke completes

- Updated the CLI Node support floor to Node 20
  - `packages/petdex-cli/package.json` now advertises `node >=20`
  - CLI README and site docs now describe Node 20+ support
  - CI intentionally tests Node 20, 22, and 24, matching the supported runtime floor

## Notes

This is intentionally not a full login E2E test. A full login test would need a Clerk test user or staging Clerk app and should use repository secrets, so it is better suited for a separate, explicitly scoped workflow.

This smoke test catches the #178 class of failure earlier: the browser reaches Clerk authorize, but Clerk rejects the CLI client before showing sign-in. It now uses the same dynamic loopback callback-port shape as the real CLI login path, so exact-port redirect URI regressions should fail the smoke.

Related to #178.

## Validation

Local validation run:

- `bun install --frozen-lockfile` in `packages/petdex-cli`
- `bun run build`
- `bun run typecheck`
- `bun test`
- `node dist/petdex.js --version`
- `npm pack --cache /private/tmp/petdex-npm-cache --silent --pack-destination /private/tmp/petdex-cli-pack-node20-check-232`
- `npx --cache /private/tmp/petdex-npm-cache -y --package /private/tmp/petdex-cli-pack-node20-check-232/petdex-0.3.9.tgz petdex --version`
- `node scripts/oauth-authorize-smoke.mjs`
- `git diff --check`
- YAML syntax parse for the workflow files
- `biome check packages/petdex-cli/scripts/oauth-authorize-smoke.mjs`

The OAuth smoke reached Clerk authorize with a dynamically assigned loopback callback port and without `invalid_client` during local validation.

### DIFF
--- a/.github/workflows/cli-auth-smoke.yml
+++ b/.github/workflows/cli-auth-smoke.yml
@@ -1,0 +1,36 @@
+name: cli-auth-smoke
+
+on:
+  schedule:
+    - cron: "17 */6 * * *"
+  push:
+    branches: [main]
+    paths:
+      - ".github/workflows/cli-auth-smoke.yml"
+      - "packages/petdex-cli/**"
+      - "src/app/api/cli/**"
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  oauth-authorize:
+    name: Production OAuth authorize smoke
+    runs-on: ubuntu-latest
+    timeout-minutes: 3
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Setup Node
+        uses: actions/setup-node@v4
+        with:
+          node-version: 24
+
+      - name: Validate CLI OAuth client
+        working-directory: packages/petdex-cli
+        # This stops before any user sign-in. It only verifies that Clerk accepts
+        # the public CLI OAuth client far enough to avoid invalid_client.
+        run: node scripts/oauth-authorize-smoke.mjs
--- a/.github/workflows/cli-ci.yml
+++ b/.github/workflows/cli-ci.yml
@@ -24,7 +24,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
-        node: [20, 22, 24]
+        node: [18, 20, 22, 24]
 
     defaults:
       run:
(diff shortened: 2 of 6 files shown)


Railly:
Nice work @iambinlin!

PR #242, opened 2026-05-13
perf(isr): hoist metrics off SSR shells + tag-based invalidation

camilocbarrera:
## Summary

Vercel bill last cycle was \$208 — dominated by ISR writes (8.83M / \$35), Active CPU + Function Invocations + Fast Origin Transfer (~\$85 combined).

Per [Vercel's ISR pricing docs](https://vercel.com/docs/incremental-static-regeneration/limits-and-pricing): *"When revalidation runs and the content hasn't changed from the previous version, no ISR write units are incurred."* Our pet pages baked install/download/like counters straight into the SSR shell, so every metric tick produced a different byte-string and burned a write. Combined with `revalidate = 60` on most pages, this caused the write storm.

This PR addresses it with three coordinated changes.

## Phase 1 — Hoist volatile data off SSR shells

- New endpoint `src/app/api/pets/[slug]/metrics/route.ts` — returns counters + global summary, same-origin gated, IP-rate-limited (240/hr), CDN-cached (`s-maxage=60, stale-while-revalidate=300`).
- New client components: `PetCountersBar` (install/download text) and `PetRadarClient` (radar with client-side stat compute). Both fetch the new endpoint and skeleton briefly on first paint.
- `LikeButton` drops the `initialCount` prop; it fetches its own count (anon via `/metrics`, signed-in via `/like` for the liked state).
- `/pets/[slug]` page strips `getMetricsForSlug` + `getMetricsSummary` from the SSR `Promise.all`, drops the JSON-LD `interactionStatistic` block (was \`metrics.likeCount\`), and removes \`stats\` computation. Shell is now byte-stable between editorial edits.

## Phase 2 — Long ISR windows

Bumped \`revalidate\` ceilings:

| Page | Before | After |
|---|---|---|
| /pets/[slug] | 60s | 86400s |
| /[locale] (home) | 60s | 86400s |
| /collections | 300s | 86400s |
| /kind/[kind] | 600s | 86400s |
| /vibe/[vibe] | 600s | 86400s |
| sitemap.ts | 3600s | 86400s |

## Phase 3 — Tag-based on-demand invalidation

Wrapped key fetches in \`unstable_cache\` with tags so write paths can flush instantly via \`revalidateTag\`:

- \`getPet(slug)\` → \`pet:\${slug}\`, \`pet:list\`
- \`getApprovedPetsWithMetrics\`, \`getFeaturedPetsWithMetrics\`, \`getAllApprovedPets\` → \`pet:list\`
- \`getCollection(slug)\` → \`collection:\${slug}\`, \`collection:list\`
- \`getCollectionsBySlugs\`, \`getCollectionsForListing\`, \`getAllCollections\` → \`collection:list\`

Calls to \`revalidateTag(tag, 'max')\` (Next 16's stale-while-revalidate signature) added in four endpoints that change displayed pet content:

- \`api/admin/[id]/feature/route.ts\` — toggle featured
- \`api/admin/edits/[id]/route.ts\` — approve queued edit
- \`api/my-pets/claim/route.ts\` — claim ownership
- \`api/my-pets/[id]/edit/route.ts\` — auto-approve branch

Skipped: submit/withdraw (status=pending pets don't appear on listings), reject/queue branches (pending fields only — no displayed change), gallery-order (only profile pages).

## Expected cost impact

| Line | Before | Target |
|---|---|---|
| ISR Writes | \$35.33 | <\$1 |
| Active CPU | \$23.70 | \$4–6 |
| Function Invocations | \$20.76 | \$4–6 |
| Fast Origin Transfer | \$40.11 | \$7–10 |
| **Subtotal** | **\$120** | **\$15–25** |

Edge Requests (\$46) and Web Analytics (\$15) are not addressed here — left for follow-up (AI crawler wall in proxy.ts, drop Analytics if not used). SpeedInsights kept intact for before/after Web Vitals comparison.

## What's deliberately left for follow-up

- \`/collections/[slug]\` is still \`force-dynamic\` — per-user \`caughtSlugs\` + auth needs Tier-1-style refactor.
- \`PetCard\` SSR'd metrics on listing pages (home/kind/vibe). At 86400s ceilings the residual writes are bounded to a handful per page per day, so it's no longer a fire — but a full refactor would zero it out.

## Test plan

- [ ] \`bun run dev:mock\` and load /en/pets/<slug>: counters flash "—" briefly then hydrate; radar fills after fetch; like button shows "—" then real count.
- [ ] Network tab: one GET /api/pets/<slug>/metrics with \`Cache-Control: public, s-maxage=60, stale-while-revalidate=300\`.
- [ ] Owner-edit a pet via My Pets (auto-approve path) — change shows up on /pets/<slug> on next visit (stale-while-revalidate).
- [ ] Admin: flip a pet's featured flag — appears on home page after next visit.
- [ ] Anon visitor sees correct like count on first paint (after small hydration delay).
- [ ] Rate limit kicks in for direct /api/pets/<slug>/metrics flood (240/hr per IP).
- [ ] Watch Vercel usage dashboard for a few days post-deploy to confirm ISR writes collapse.

### DIFF
--- a/src/app/[locale]/collections/page.tsx
+++ b/src/app/[locale]/collections/page.tsx
@@ -12,9 +12,10 @@ import { SiteHeader } from "@/components/site-header";
 import { hasLocale } from "@/i18n/config";
 
 // /collections is fully public — no auth, no cookies, no per-visitor
-// data. ISR with 5 minute revalidation lets featured-collection
-// rotations show up quickly without waking a function on every visit.
-export const revalidate = 300;
+// data. 24h ceiling + revalidateTag('collection:list') from admin
+// write paths keeps the page fresh on actual changes without burning
+// a function on every visit.
+export const revalidate = 86400;
 
 const SITE_URL = "https://petdex.crafter.run";
 const MIN_PETS = 4;
--- a/src/app/[locale]/kind/[kind]/page.tsx
+++ b/src/app/[locale]/kind/[kind]/page.tsx
@@ -14,7 +14,7 @@ const SITE_URL = "https://petdex.crafter.run";
 
 type Props = { params: Promise<{ kind: string; locale: string }> };
 
-export const revalidate = 600;
+export const revalidate = 86400;
 
 export function generateStaticParams() {
   return PET_KINDS.map((kind) => ({ kind }));
--- a/src/app/[locale]/page.tsx
+++ b/src/app/[locale]/page.tsx
@@ -36,15 +36,14 @@ import {
 
 import { hasLocale, locales } from "@/i18n/config";
 
-// ISR. The home page used to be force-dynamic because it pulled the
-// visitor's shuffle seed cookie and caught-slug set. Both moved to
-// the client: PetGallery re-fetches /api/pets/search after hydration
-// (which picks up the cookie) and /api/me/caught-slugs feeds the
-// "caught" highlight. The server now renders an alpha-ordered, anon
-// shell that the edge can cache for 60s — enough to keep new pets
-// surfacing without waking a function on every visit.
+// ISR. The home page renders an alpha-ordered, anon shell — the
+// visitor's shuffle seed and caught-slug set are pulled client-side
+// (PetGallery re-fetches /api/pets/search; /api/me/caught-slugs feeds
+// the "caught" highlight). With a 24h ceiling and tag-based
+// invalidation on submit/feature/withdraw, the page stays fresh for
+// editorial changes without burning a function on every visit.
 export const dynamic = "force-static";
-export const revalidate = 60;
+export const revalidate = 86400;
 
 export async function generateMetadata({
   params,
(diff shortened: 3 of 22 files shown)


iambinlin:
Cris, one follow-up idea: adding hotlink protection for the public R2 assets could also be a good way to save cost.

I just tested a `r2.dev/pets/.../sprite.webp` style asset, and it returns 200/206 with no Referer, with an external Referer, and with an external Origin. That means if another site hotlinks these images directly, the R2/Cloudflare traffic cost can still land on us.

For a follow-up, we could consider serving the assets from a custom domain and adding a Cloudflare WAF/Rules allowlist based on Referer. For the highest-traffic assets, signed URLs or short-lived tokens would be stronger. This probably does not need to be part of the current ISR PR, but it is in the same cost-reduction family.

PR #274, opened 2026-05-16
feat: add visual OCR submission review

Railly:
## Summary
- review a contact sheet from submitted spritesheet frames instead of only the first frame, with a neutral background for OCR visibility
- use the same fixed 192x208 / 1536x1872 sprite grid as rendering, and fail closed for non-ideal decoded dimensions instead of sampling shifted pixels
- skip unused per-state frame cells via petStates so hidden template padding cannot trigger policy holds
- render the policy contact sheet from a single decoded raw buffer, with one PNG encode, to keep review cost bounded
- require OCR-oriented policy review for embedded text, symbols, likenesses, historical/religious figures, and other visual risk signals
- persist visual text and visual signal evidence in submission review checks and show it in the admin review evidence panel
- localize the admin evidence summary/fallback/confidence labels in en, es, and zh
- surface specific manual-review reasons when image preparation cannot run, including non-ideal dimensions and model payload limits
- keep text-only edit moderation on a non-OCR prompt so image-review instructions do not regress edit auto-accept behavior
- fail closed if edit moderation returns any flags, even with decision pass
- tolerate valid JSON wrapped in model prose while still failing closed when no valid JSON exists
- bound asset fetches and policy model calls so automated review work cannot outlive the submission review timeout indefinitely
- harden malformed policy responses so risky, invalid, or unknown-category model output fails closed into manual review instead of being silently dropped
- cap source/model payload size and redact arbitrary pet.json metadata before policy review

Closes #238

## Checks
- bun test src/lib/edit-policy.test.ts src/lib/submission-review.test.ts: 30 pass
- bun x biome check src/lib/submission-review.ts src/lib/submission-review.test.ts
- bun x biome check src/lib/submission-review-image.ts src/lib/submission-review.test.ts
- bun run i18n:check
- bun x biome check src/components/admin-review-row.tsx src/i18n/messages/en.json src/i18n/messages/es.json src/i18n/messages/zh.json
- bun run build

## Review notes
- Latest SHA: 50f9a31cdcaabb7a6aacd7c833e3633ae022db42.
- PR #275 is merged into main, so GitHub opencode-review runs on this PR with the repaired workflow.
- GitHub opencode-review review and publish checks pass on the latest SHA.
- Latest opencode comment on the latest SHA says: No actionable findings.
- Vercel preview check passes on the latest SHA.
- Build still emits the existing Turbopack NFT and unstable_cache >2MB warnings outside this PR.
[...]

PR #275, opened 2026-05-16
fix: repair opencode review workflows

Railly:
## Summary
- stop passing a synthetic workflow_dispatch payload to opencode review
- run the automatic pull_request_target review with opencode run and publish its returned body once
- pass the GitHub token directly to the comment-triggered opencode action to avoid the failing app token exchange path

## Checks
- ruby YAML parse for .github/workflows/opencode-review.yml and .github/workflows/opencode.yml

This is needed before #274 can get a clean opencode review check, because pull_request_target evaluates the workflow from main.

### DIFF
--- a/.github/workflows/opencode-review.yml
+++ b/.github/workflows/opencode-review.yml
@@ -8,10 +8,9 @@ jobs:
   review:
     runs-on: ubuntu-latest
     permissions:
-      id-token: write
       contents: read
-      pull-requests: write
-      issues: write
+      pull-requests: read
+      issues: read
     steps:
       - uses: actions/checkout@v6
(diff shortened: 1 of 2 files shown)


Railly:
opencode latest local review on cf57d27:

- Initial pass found two security issues: write-scoped token exposed to the pull_request_target agent, and unused id-token permission.
- Fixed by splitting automatic review into a read-only review job plus separate publish job with write token, removing unused OIDC, and restricting comment-triggered `/opencode` to OWNER/MEMBER/COLLABORATOR before passing `TOKEN`.
- Final opencode pass: no actionable findings.

PR #294, opened 2026-05-20
fix(ci): use --no-bundle flag for tauri-cli v2 and add prerelease detection

boygotflames:
## Problem

`cargo tauri build --bundles none` is invalid in tauri-cli v2.
Valid flags are `msi`, `nsis`, or `--no-bundle`.
Every `desktop-v*` tag pushed before this fix would fail immediately
on the Windows build step, producing no artifact.

## Changes

- `build-windows` job: `--bundles none` → `--no-bundle`
- Release step: add `prerelease: ${{ contains(github.ref_name, '-rc') }}`
  so rc tags publish as prerelease automatically

## Verification

Local build confirmed with `cargo tauri build --no-bundle`:
- Binary: `petdex-desktop-win32-x64.exe`
- Size: 8.33 MB
- Artifact name matches workflow expectation exactly

## Notes

The macOS build job is untouched. This only affects the `build-windows` job.

When ready to test, pushing a `desktop-v*-rc.1` tag will exercise
the full CI pipeline as prerelease before any stable tag is created.

### DIFF
--- a/.github/workflows/desktop-release.yml
+++ b/.github/workflows/desktop-release.yml
@@ -178,3 +178,4 @@ jobs:
           files: release/*
           generate_release_notes: true
           fail_on_unmatched_files: true
+          prerelease: ${{ contains(inputs.tag || github.ref_name, '-rc') }}
--- a/.github/workflows/desktop-release.yml
+++ b/.github/workflows/desktop-release.yml
@@ -122,12 +122,13 @@ jobs:
         run: cargo install tauri-cli --version "^2" --locked
 
       - name: Build petdex-desktop-win32-x64.exe
-        # `--bundles none` skips NSIS/WiX installer generation (we ship the
-        # bare .exe per the CLI asset contract). The binary name is driven by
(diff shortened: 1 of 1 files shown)


boygotflames:
Hi @Railly  — the `build-windows` job is confirmed in `main` and a local
`cargo tauri build --no-bundle` produces `petdex-desktop-win32-x64.exe` at
8.33 MB cleanly in 49s.

And yeah, One fix was needed: `--bundles none` is not a valid flag in tauri-cli v2
(only `msi` and `nsis` are accepted). The correct flag is `--no-bundle`.
PR #294 fixes this in the workflow so CI does not fail on first trigger.

Would a prerelease tag (e.g. `desktop-v0.3.0-rc.1`) work for the first
Windows build test, so we can verify the artifact before marking it stable?
The workflow also supports `workflow_dispatch` with a `tag` input if you
prefer to trigger it manually without pushing a git tag.

Trigger URL: https://github.com/crafter-station/petdex/actions/workflows/desktop-release.yml
Input: `tag = "desktop-v0.3.0-rc.1"` (or whatever next version)

im ready, to procced with work! >>>

PR #362, opened 2026-06-04
Slim collections page payload

Railly:
## Summary
- pass only the cover pet fields needed by the collections browser client component
- keep the existing collection listing UI and cache behavior unchanged
- reduce serialized `/collections` payload pressure behind Fast Origin Transfer

## Evidence
- production `/collections` is already cache HIT, but transfers about 1.04 MB uncompressed and 180 KB gzip
- the previous client boundary received full `PetWithMetrics` objects for each preview pet, while the cover only reads `slug`, `displayName`, and `spritesheetPath`

## Validation
- `bunx biome check --write src/app/[locale]/collections/page.tsx src/components/collection-cover.tsx src/components/collections-browser.tsx`
- `bun run check`
- `bun run i18n:check`
- `git diff --check`
- `bun test --env-file=.env.mock`
- `TELEMETRY_RATELIMIT_SECRET=mock-telemetry-secret bun --env-file=.env.mock run build`

### DIFF
--- a/src/app/[locale]/collections/page.tsx
+++ b/src/app/[locale]/collections/page.tsx
@@ -89,7 +89,11 @@ export default async function CollectionsPage({
     externalUrl: c.externalUrl,
     coverPetSlug: c.coverPetSlug,
     petCount: c.petCount,
-    pets: c.pets,
+    pets: c.pets.map((pet) => ({
+      slug: pet.slug,
+      displayName: pet.displayName,
+      spritesheetPath: pet.spritesheetPath,
+    })),
   }));
 
   return (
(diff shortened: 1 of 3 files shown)


Railly:
Measurement update:

- Preview URL returns `401` to unauthenticated curl because Deployment Protection is enabled, so I could not measure preview transfer directly.
- Production baseline for `/collections`: `1,040,067` bytes plain, `180,283` bytes gzip, `x-vercel-cache: HIT`.
- Local production build from this PR with `.env.local`: `151,368` bytes plain, `40,179` bytes gzip, `x-nextjs-cache: HIT`.
- Expected reduction: about `85%` plain and `78%` gzip for this route payload, without changing route cache behavior.

Issue #463, opened 2026-06-12
Download pet fail


henryjing96:
## Bug

`npx petdex@latest install 11` fails when downloading pet assets from `assets.petdex.dev`.

## Reproduction

```bash
npx petdex@latest install 11
```

Observed output:

```text
Failed:
  11: download https://assets.petdex.dev/pets/11-8115f64f9fbf/petjson.json -> 403 Forbidden
```

<img width="2212" height="594" alt="Image" src="https://github.com/user-attachments/assets/b8386dcc-6304-4640-bffd-68697deaaeb1" />

## Expected behavior

The pet should install successfully into:

- `~/.petdex/pets/11`
- `~/.codex/pets/11`

## What I found

The npm `latest` package is currently `petdex@0.4.3`.

The published package still defaults to the legacy base URL:

```text
https://petdex.crafter.run
```

That means asset downloads send:

```text
Referer: https://petdex.crafter.run/
```

But the current `assets.petdex.dev` Cloudflare/WAF hotlink policy allows `petdex.dev`, not the legacy `petdex.crafter.run` referer, so the request is blocked with `403`.

The same install succeeds when overriding the base URL:

```bash
PETDEX_URL=https://petdex.dev npx petdex@latest install 11
```

## Suggested fix

Publish a new CLI version, for example `petdex@0.4.4`, built from the current source where the default base URL is:

```ts
const PETDEX_URL = process.env.PETDEX_URL ?? "https://petdex.dev";
```

Also bump the CLI version constant so `petdex --version` matches the npm package version.

## Validation

After publishing, this should succeed without any environment override:

```bash
tmp_home="$(mktemp -d)"
HOME="$tmp_home" npx -y petdex@latest install 11
find "$tmp_home" -maxdepth 5 -type f | sort
rm -rf "$tmp_home"
```

Expected files:

```text
~/.codex/pets/11/pet.json
~/.codex/pets/11/spritesheet.webp
~/.petdex/pets/11/pet.json
~/.petdex/pets/11/spritesheet.webp
```

PR #464, opened 2026-06-12
fix(cli): release petdex.dev URL patch

henryjing96:
## Summary

- bump the CLI package and runtime version to 0.4.4 for a new npm release
- keep the production CLI default pointed at https://petdex.dev
- add a release contract test so package version, CLI version, and the production URL do not drift again

## Root cause

The published `petdex@0.4.3` package still defaulted to the legacy `https://petdex.crafter.run` base URL. Asset downloads therefore sent `Referer: https://petdex.crafter.run/`, which is blocked by the current `assets.petdex.dev` hotlink/WAF policy. Overriding `PETDEX_URL=https://petdex.dev` makes the install succeed, so the fix is to publish a fresh CLI package with the canonical default.

## Validation

- `bun run build`
- `bun run typecheck`
- `bun test`
- `node dist/petdex.js --version` -> `0.4.4`
- `npm pack --silent` -> `petdex-0.4.4.tgz`
- `rg "petdex\\.crafter\\.run" dist/petdex.js` -> no matches
- temporary-HOME smoke: `PETDEX_TELEMETRY=0 HOME="$tmp_home" node ./dist/petdex.js install 11` wrote both `.petdex/pets/11` and `.codex/pets/11`

## Release handoff

After merge, an npm maintainer should publish from `packages/petdex-cli`:

```bash
bun install --frozen-lockfile
bun run build
bun run typecheck
bun test
npm publish --access public
npm view petdex version
```

Then verify npm latest without an env override:

```bash
tmp_home="$(mktemp -d)"
HOME="$tmp_home" npx -y petdex@latest install 11
rm -rf "$tmp_home"
```

### DIFF
--- a/packages/petdex-cli/bin/petdex.ts
+++ b/packages/petdex-cli/bin/petdex.ts
@@ -114,7 +114,7 @@ async function getAuth(): Promise<ClerkCliAuth> {
   return _auth;
 }
 
-const VERSION = "0.4.2";
+const VERSION = "0.4.4";
 
 // ─── entrypoint ────────────────────────────────────────────────────────────
 main().catch((err) => {
--- a/packages/petdex-cli/package.json
+++ b/packages/petdex-cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "petdex",
-  "version": "0.4.2",
+  "version": "0.4.4",
   "description": "CLI for Petdex. Install, browse, and submit Codex pets from your terminal.",
   "homepage": "https://petdex.dev",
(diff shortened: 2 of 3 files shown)


henryjing96:
This PR addresses the issue reported in #463: petdex@0.4.3 installs can fail with a 403 because the published CLI still uses the legacy petdex.crafter.run referer for asset downloads.

henryjing96:
fix in https://github.com/crafter-station/petdex/pull/501

PR #466, opened 2026-06-12
fix: prevent r2 presigned put checksum mismatch

### LINKED ISSUE #465: R2 presigned PUT upload fails during pet submission
## Symptom

When submitting a pet from the web `/submit` page, validation passes and the UI shows:

> Looks ready to submit.

After clicking **Submit pet**, the upload fails with:

```text
Upload failed: R2 PUT network error: xhr network error (size=290, type=application/json)
```

Image

The failure happens after the app receives R2 presigned URLs, during the browser's first direct PUT upload of `pet.json`.

## Root Cause

This is not a normal network failure. CSP and R2 CORS preflight are already allowing the upload request.

The R2 presigned PUT URL is generated by AWS SDK v3. Newer SDK behavior adds flexible checksum query params to `PutObject` presigned URLs, such as:

```text
x-amz-checksum-crc32=AAAAAA==
x-amz-sdk-checksum-algorithm=CRC32
```

`AAAAAA==` is the CRC32 checksum for an empty payload. The server generates the presigned URL without the real upload body, but the browser later PUTs a non-empty `pet.json`.

R2 rejects the request because the signed checksum does not match the uploaded body. Since the R2 error response is not readable by the browser as a CORS response, XHR only reports the generic `xhr network error`.

## Proposed Fix

Configure the shared R2 S3 client to calculate request checksums only when the service requires them:

```ts
export const r2 = new S3Client({
  region: "auto",
  endpoint: `https://${ACCOUNT_ID ?? "missing"}.r2.cloudflarestorage.com`,
  requestChecksumCalculation: "WHEN_REQUIRED",
  credentials: {
    accessKeyId: ACCESS_KEY_ID ?? "",
    secretAccessKey: SECRET_ACCESS_KEY ?? "",
  },
});
```

Add a regression test for `presignPut()` to ensure the generated URL:

- still signs the `content-type` header
- uses `UNSIGNED-PAYLOAD`
- does not include `x-amz-checksum-crc32`
- does not include `x-amz-sdk-checksum-algorithm`

## Verification

```bash
bun test src/lib/r2.test.ts
bun run check
TELEMETRY_RATELIMIT_SECRET=petdex-build-check-secret bun --env-file=.env.mock run build
```

## Impact

This affects R2 direct-upload flows that reuse `presignPut()`, including:

- web pet submission
- CLI submission presign
- owner edit presign
- ad image uploads
- pet request image uploads

The fix is centralized in `src/lib/r2.ts` and should be low risk.

henryjing96:
## Summary

Fixes crafter-station/petdex#465

This PR applies an analysis-driven fix for the `/submit` upload failure where browser-side R2 PUT requests surfaced as a generic `xhr network error`.

The investigation found that AWS SDK v3 was adding flexible checksum query parameters to presigned `PutObject` URLs. In this flow, the server signs the URL without the real browser upload body, so the generated URL can include an empty-payload checksum. When the browser later uploads a non-empty `pet.json`, R2 can reject the request, and the browser reports it as a generic XHR network error.

This fix comes from the issue analysis result.

## Changes

- Configure the shared R2 `S3Client` with `requestChecksumCalculation: "WHEN_REQUIRED"`.
- Add a regression test for `presignPut()` to ensure generated URLs:
  - still sign `content-type`
  - use `UNSIGNED-PAYLOAD`
  - do not include `x-amz-checksum-crc32`
  - do not include `x-amz-sdk-checksum-algorithm`

## Verification

Passed locally:

- `bun test src/lib/r2.test.ts`
- `bun run check`
- `TELEMETRY_RATELIMIT_SECRET=petdex-build-check-secret bun --env-file=.env.mock run build`

## Validation Status

This has not yet completed full end-to-end verification against the real deployed `/submit` flow.

Real R2 upload validation was not performed locally because the environment does not have a confirmed usable Cloudflare API token or real R2 write credentials.

Required follow-up validation after deploy:

- Open `/submit`.
- Upload a valid pet pack.
- Confirm `pet.json`, sprite, and zip R2 PUT requests succeed.
- Confirm the submission is created and enters pending state.

### DIFF
--- a/src/lib/r2.test.ts
+++ b/src/lib/r2.test.ts
@@ -1,0 +1,26 @@
+import { describe, expect, it } from "bun:test";
+
+describe("presignPut", () => {
+  it("does not sign an empty payload checksum for browser uploads", async () => {
+    process.env.R2_ACCOUNT_ID = "62819ee0a8411123c2635cbf37b577c1";
+    process.env.R2_ACCESS_KEY_ID = "test";
+    process.env.R2_SECRET_ACCESS_KEY = "test";
+    process.env.R2_BUCKET = "petdex-pets";
+
+    const { presignPut } = await import("./r2");
+    const presigned = await presignPut(
+      "pets/test-upload/petjson.json",
+      "application/json",
+    );
+
+    const url = new URL(presigned.uploadUrl);
+    expect(url.searchParams.get("X-Amz-SignedHeaders")).toContain(
+      "content-type",
+    );
+    expect(url.searchParams.get("X-Amz-Content-Sha256")).toBe(
+      "UNSIGNED-PAYLOAD",
+    );
+    expect(url.searchParams.has("x-amz-checksum-crc32")).toBe(false);
+    expect(url.searchParams.has("x-amz-sdk-checksum-algorithm")).toBe(false);
+  });
+});
--- a/src/lib/r2.ts
+++ b/src/lib/r2.ts
@@ -22,6 +22,7 @@ if (!ACCOUNT_ID || !ACCESS_KEY_ID || !SECRET_ACCESS_KEY) {
 export const r2 = new S3Client({
   region: "auto",
   endpoint: `https://${ACCOUNT_ID ?? "missing"}.r2.cloudflarestorage.com`,
+  requestChecksumCalculation: "WHEN_REQUIRED",
   credentials: {
     accessKeyId: ACCESS_KEY_ID ?? "",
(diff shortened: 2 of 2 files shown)


Railly:
Thanks @henryjing96 — same root cause and fix you nailed in #465. This shipped in #475 (merged) with the centralized `requestChecksumCalculation: "WHEN_REQUIRED"` on the shared R2 client plus a regression test. Closing this as a duplicate of that merge. Appreciate the diagnosis 🙏

PR #495, opened 2026-06-21
fix: accept single-pet folder zip uploads

henryjing96:
## Summary
- Support ZIP uploads where `pet.json` and `spritesheet.webp/png` are inside a single top-level pet folder.
- Prefer root-level pet files when both root and nested pet structures exist.
- Ignore macOS/system metadata files during ZIP classification.

## Validation
- `bun test src/lib/pet-zip.test.ts`
- `bun run check`
- `TELEMETRY_RATELIMIT_SECRET=mock-build-secret bun --env-file=.env.mock run build`


henryjing96:
no need to support this feature

Issue #492, opened 2026-06-17
petdex-desktop segfaults on launch (Intel x86_64, macOS 26 Tahoe) in -[ZeroNativeBorderlessWindow canBecomeKeyWindow]


lwma624-netizen:
petdex-desktop segfaults on launch (Intel x86_64, macOS 26 Tahoe) in -[ZeroNativeBorderlessWindow canBecomeKeyWindow]

Summary
petdex-desktop crashes with a segmentation fault immediately on startup, while creating the mascot window. It never shows a window. The crash is 100% reproducible and independent of which pet is installed (an installed, valid pet aurelion-sol loads fine right before the crash).

Environment
Hardware: Intel x86_64 Mac
OS: macOS 26.3.1 (Tahoe), build 25D771280a
petdex CLI: 0.4.2 (also reproduced via npx petdex@latest)
Desktop binary: desktop-v0.2.2 (petdex update reports "Already up to date")
Binary arch confirmed matching: file petdex-desktop → Mach-O 64-bit executable x86_64 (not a Rosetta/arch mismatch)
petdex doctor is otherwise green: desktop binary, sidecar bundle, installed pets, and all agent hooks (Claude Code, Codex, OpenCode, Antigravity) are installed correctly. Only the desktop window fails.

Steps to reproduce
On an Intel Mac running macOS 26 (Tahoe), install a pet: npx petdex install aurelion-sol
Launch the mascot: petdex up (or petdex desktop start, or run the binary directly)
Check ~/.petdex/desktop.log
Expected
The mascot window appears.

Actual
Segmentation fault during window creation; the process crashes (and crash-loops if started repeatedly).

Crash log (~/.petdex/desktop.log)
Loading pet: aurelion-sol (1 installed)
Spritesheets: 0 copied, 1 cached
petdex: sidecar spawned (node /Users/edy/.petdex/sidecar/server.js)
Segmentation fault at address 0x0
???:?:?: 0x...de0 in -[ZeroNativeBorderlessWindow canBecomeKeyWindow] (/Users/edy/.petdex/bin/petdex-desktop)
???:?:?: 0x... in ??? (/System/Library/Frameworks/AppKit.framework/Versions/C/AppKit)
???:?:?: 0x... in ??? (/System/Library/Frameworks/AppKit.framework/Versions/C/AppKit)
[...]

PR #519, opened 2026-06-28
feat(desktop): add --direct launch mode for macOS app bundle workaround

reddoctor:


henryjing96:
Thanks, this looks like the right conservative workaround.

Merging because `petdex desktop start --direct` gives affected Intel/macOS Tahoe users a CLI path that bypasses the `open -gj` LaunchServices launch path, while preserving the default `petdex desktop start` behavior for everyone else.

This should be treated as a workaround, not the root fix for the zero-native/AppKit borderless window crash. We should continue tracking the native crash separately in #492/#515 and ask affected users to try:

```bash
petdex desktop stop
petdex desktop start --direct

PR #578, opened 2026-07-19
feat: native desktop app rewrite on native-sdk

Railly:
Full rewrite of the desktop pet on the native-sdk UI engine, replacing the WebView app. Supersedes the approach in the original desktop package; cutover is planned as a later major once this stabilizes.

- Pet window sized exactly to the sprite: states, drag with throw physics, context menu, native menu bar
- Speech bubbles in their own fixed-size floating click-through window: session title, activity preview, agent avatar, busy spinner
- In-binary hook runner behind a stable ~/.petdex/bin/petdex-hook symlink, so the hook hot path costs native startup instead of a node cold start
- Settings: pet picker with search, size slider, message toggle, and an Agents section with hook detection plus one-click install/update/disconnect for Claude Code, Codex, Gemini and opencode (non-destructive JSON merges, one-time backups)
- opencode plugin ported to the current function-export API
- Builds against a vendored framework overlay carrying the floating/transparent/click-through window flags and the moveWindow/resizeWindow verbs; draft until that lands upstream

https://claude.ai/code/session_01DXYbEzfNkXudCrL2Adcmqq


Railly:
Two commits on top of the rewrite, plus what was verified running the packaged app.

## `55accb0` agent logos were invisible in Settings

Both the Settings agent rows and the bubble avatar opened `assets/agents/<name>.png` relative to the process cwd. That holds under `zig build`, but `open`ing a .app starts it with cwd `/`, so each probe resolved to `/assets/...` and the loop hit its `continue` four times over.

Confirmed before the fix: `lsof -a -p <pid> -d cwd` reported `/`, and no `/tmp/petdex-native-agenticon-*.tga` files existed at all, meaning no PNG was ever found.

`assetPath` now resolves against `_NSGetExecutablePath`, probing `Contents/Resources` first and the directory beside the binary second, so the packaged app and a plain `zig build` run both work. Same approach `packages/petdex-desktop` already uses for its sidecar.

After: all four TGAs appear and the logos render beside each agent name.

## `b10c3bb` the native app had no release path

`desktop-release.yml` only ever built the WebView desktop. There is now a `build-native` job beside it, so a `desktop-v*` tag produces both. The existing job is untouched.

Two things worth a reviewer's eye:

- The SDK comes from `Railly/native:feat/floating-window-0.5.3`, pinned to `b414704d`, because `floating` / `transparent` / `click_through` / `fx.focusWindow` are not in `vercel-labs/main` yet.
- The CLI is built from that same commit rather than installed from npm. The published `@native-sdk/cli` carries its own manifest parser, and the released one rejects `app.zon`'s `.floating` with a bare `ParseZon` and no field name. That is the whole reason this branch appeared not to build.

Both new steps were run locally against the packaged app: the logo assert passes and `ditto` produces a 3.2MB zip.

## Verified running

Packaged with adhoc signing and launched. `CGWindowList` reports `cgLayer=3` (`NSFloatingWindowLevel`), onscreen, 98x107 — always-on-top and transparency both real, with the terminal readable behind the sprite. The activity bubble tracked live tool calls, so the agent hooks are working end to end.

## Not resolved

The release still depends on an unpublished SDK fork. Anyone cloning this and running `native build` with the npm CLI hits the same `ParseZon`. Worth deciding whether to document the build-from-fork step in the package README, or wait for those options to land upstream.

Issue #585, opened 2026-07-22
宠物封面显示不出来


2204160415:
宠物编号4004，名称“沈青竹”，7月21日上传的，封面显示不出来

<img width="1571" height="1244" alt="Image" src="https://github.com/user-attachments/assets/c4014ae2-658a-482d-9855-e6e888059892" />

Railly:
Fixed by #587 + the preview backfill that ran right after merging (320 missing previews regenerated, cache purged). Your pet's cover should render now — and even if a preview is ever missing again, cards now fall back to the full spritesheet instead of going blank. Please reopen if you still see it blank after a hard refresh.

PR #587, opened 2026-07-23
fix(gallery): fall back to atlas when a card preview is missing, unbreak preview backfill

### LINKED ISSUE #579: Creator profile card preview is blank while pet detail page renders correctly
## Description

The pet preview is blank on the creator profile card, while the same pet renders correctly on its detail page and in the state viewer.

Pet URL: https://petdex.dev/pets/lingxia

## Steps to reproduce

1. Open the creator profile page that lists the submitted pet.
2. Observe the "明日香" pet card.
3. The preview area is blank.
4. Open the pet detail page.
5. The hero animation and individual animation states render correctly.

## Expected behavior

The creator profile card should display a preview of the pet, consistent with the detail page.

## Actual behavior

Only the profile card preview is blank. Metadata, detail-page animation, and state thumbnails all load correctly.

## Asset information

- Format: WebP with transparency
- Size: 1536 × 2288
- Grid: 8 columns × 11 rows
- Metadata: `spriteVersionNumber: 2`

This suggests a profile-card thumbnail/cache/rendering issue rather than an invalid uploaded spritesheet.

Jibaru:
Closes #579

## Problem

The `lingxia` pet card is blank on its creator profile while the detail page renders fine. Three layers to it:

1. **Blank card** — profile cards ([`PetCard`](../blob/main/src/components/pets/pet-gallery.tsx)) and collection covers render a derived `pets/<slug>/preview.webp` in `layout="row"`, unlike the detail page which uses the full atlas. The URL is emitted without any existence check, and a CSS `background-image` 404s silently → blank. `https://assets.petdex.dev/pets/lingxia/preview.webp` is a genuine 404 (`cf-cache-status: DYNAMIC`, so the object is absent from R2, not a stale CDN entry).
2. **Why it stayed missing** — the 6-hourly `preview-snapshot.yml` backfill that reconciles missing previews has failed on **every run since 2026-07-18** with `error: lockfile had changes, but lockfile is frozen`.
3. **Root cause** — Dependabot #552 (`ai` `^6.0.224`→`^7.0.22`) and #538 (`next` →`16.2.10`) changed only `package.json`; `bun.lock` still pinned `ai@6.0.175` / `next@16.2.9`. Every `--frozen-lockfile` workflow on main broke (`manifest-snapshot.yml` too), and there was no PR CI to catch the drift.

## Changes

- **`bun.lock`** — regenerated in sync with `package.json` (`ai@7.0.35`, `next@16.2.10`). Revives both scheduled snapshot workflows.
- **`PetSprite`** — new optional `fallbackSrc` prop: an invisible 1px `` probes the preview URL (shares the browser cache with the background fetch, so the happy path costs no extra request) and swaps the card to the atlas on load error. Wired at the three preview call sites (`pet-gallery.tsx`, `collection-cover.tsx` ×2). Error-only, so it avoids the mass runtime fallback that regressed the first preview rollout — cards with a healthy preview are untouched.
- **`.github/workflows/lockfile-check.yml`** — `bun install --frozen-lockfile` on PRs/pushes so the next lockfile-less bump fails at PR time instead of silently killing cron jobs for days.
- **`.env.mock`** — mock `TELEMETRY_RATELIMIT_SECRET` (≥16 chars) so the documented `.env.mock` build-check path works again; `/api/telemetry/event` throws at import time in production builds without it.
- **`packages/petdex-cli/bin/petdex.ts`** — pre-existing Biome import-order error fixed so root `bun run check` is green.

## Verification

- `bun run check` ✅ · `bun --env-file=.env.mock run build` ✅ (validates the `ai@7` call sites: auto-tag, embeddings, pet-sound, submission-review) · app tests `src/lib src/components src/app src/i18n` **268/268** ✅
- **Fallback end-to-end**: built with `NEXT_PUBLIC_PETDEX_PET_PREVIEWS_ENABLED=true` on the mock env (where preview URLs point at unresolvable `mock.invalid`), served, and dumped the hydrated DOM via headless Chrome. SSR HTML shows cards on `…/pets/boba/preview.webp`; the hydrated DOM has **zero** `preview.webp` references and the same cards on `…/curated/boba/spritesheet.webp` — the swap that replaces the blank card.

## After merge

Dispatch the backfill once this is on main (the workflow pins `ref: main`):

```
gh workflow run preview-snapshot.yml -f mode=apply --repo crafter-station/petdex
```

It regenerates `lingxia`'s preview (and any other missing previews/thumbnails/stickers) and purges the Cloudflare cache for the uploaded URLs.

Known follow-up (out of scope): artifacts only republish on slug change, so a sprite swap under the same slug keeps a stale preview (`submission-decisions.ts`) — stale doesn't blank the card, and the new fallback also covers future deletions.


Railly:
Ran the full review gate on this locally. Everything checks out, and I verified the core claims myself rather than taking the PR body's word:

- **Lockfile force-red both ways**: main's `bun.lock` + this PR's `package.json` reproduces the exact `lockfile had changes, but lockfile is frozen` error; with the PR's lockfile, `bun install --frozen-lockfile` passes in root and in `packages/petdex-cli`.
- **Causal chain is live**: the last 3 `preview-snapshot` runs are still red, and `pets/lingxia/preview.webp` is a genuine origin 404 (`cf: EXPIRED` on revalidation, not a stale CDN entry).
- **Full suite re-run locally**: `bun run check`, 268/268 tests, mock build all green.
- **Caller sweep**: all 9 `PetSprite` call sites outside the diff pass `spritesheetPath` directly; the 3 preview consumers are exactly the ones wired with `fallbackSrc`. Coverage is complete.
- Lockfile diff is clean: only `ai` 6.0.175 -> 7.0.35, `next` 16.2.9 -> 16.2.10 and transitives.

Really good work on the diagnosis depth here, the three-layer root cause (blank card -> dead backfill -> lockfile drift) plus the error-only fallback design that avoids the first rollout's regression is exactly right.

Follow-ups filed so they don't get lost: #588 (fallback cards render fixed idle instead of cycleStates), #589 (CI test for the fallback), #590 (stale preview on same-slug sprite swap, the one you already flagged). One nit for a future pass, not blocking: the comment in `lockfile-check.yml` line 10 uses an em dash, which we keep out of added lines. Also noting #585 reports the same symptom for another pet (shenqingzhu) and should resolve with the post-merge backfill.

Merging as squash. Dispatching the backfill right after.

Railly:
Post-merge loop closed: backfill ran green — 320 missing previews uploaded (13.7 MB), cache purged. `pets/lingxia/preview.webp` now returns 200 from R2, so the card doesn't even need the fallback anymore. #585 (same symptom, different pet) closed too. Thanks again @Jibaru!

PR #591, opened 2026-07-23
feat(desktop): add a setting to hide the agent activity bubble

KoalaBear:
## Problem

The activity bubble above the pet has no off switch. The closest thing is the hooks killswitch (`/petdex` or `petdex hooks off`), but that kills the sprite state changes too — you either get the full narration or a frozen pet. During screen shares, recordings, or just a busy session, the pet reacting is the fun part and the tool-by-tool text is noise.

## Fix

A `showBubbles` preference (default `true`) in `~/.petdex/preferences.json`, next to the existing `autoInstallUpdates`, surfaced as an "Activity bubble" toggle in the desktop settings window.

Two enforcement points, on purpose:

- **`read_runtime_bubble` (main.zig)** serves the empty payload while bubbles are off. This is what makes the toggle feel immediate: an already-visible bubble hides on the next 200 ms poll instead of lingering forever (bubbles are persistent-until-replaced, so gating only the writes would pin the last bubble on screen indefinitely). `counter: 0` never collides with a real counter, so re-enabling replays the current bubble via the normal counter-change path — no JS changes needed in `pollBubble`.
- **Sidecar `POST /bubble`** skips the `bubble.json` write while suppressed. It still returns 200 (hooks are fire-and-forget and should never see an error) and still counts toward init-status, since a suppressed bubble is still proof the hooks are wired correctly.

`write_desktop_settings` previously hard-required `autoInstallUpdates` and rewrote the whole file from that one field; it now falls back to the current on-disk value for any field absent from the payload, so a future single-toggle write can't silently reset the other preference.

Local install-progress bubbles ("Installing aurora…") are intentionally unaffected — those narrate an action the user just took, not ambient hook traffic.

## Verification

- `bun test` at repo root: 469 pass (includes new `parseDesktopPreferences` cases for `showBubbles`)
- `cd packages/petdex-desktop && zig build && zig build test`: compiles, 18/18 pass (zig 0.16.0, zero-native sibling checkout per README)
- `bun run check`: clean
- `server.js` rebuilt with the sidecar's `bun run build`; the bundle loads and correctly yields EADDRINUSE against a live incumbent sidecar


Railly:
Thanks for this one, @KoalaBear.

The desktop app was rewritten as a native app and `packages/petdex-desktop`
is gone, so this diff no longer has files to land on. The setting itself
shipped: Appearance in the new app has a toggle for the activity bubble.

Grab it here: https://petdex.dev/download

PR #592, opened 2026-07-23
feat(cli): honor CLAUDE_CONFIG_DIR when wiring Claude Code hooks

KoalaBear:
## Problem

Claude Code supports fully isolated side-by-side installs through `CLAUDE_CONFIG_DIR` — separate settings, separate slash commands, even separate logged-in accounts. The agent registry hardcodes `~/.claude`, so `petdex hooks install` always detects and writes the default instance. If the Claude Code you actually launch runs from a non-default config dir, its `settings.json` never gets the hooks and the pet just sits there; there is no flag or env var to point petdex at the right place.

## Fix

`resolveClaudeConfigDir(env, home)`: prefer `CLAUDE_CONFIG_DIR`, fall back to `~/.claude` — the exact shape of the existing `resolveOpenCodeConfigDir` (which already honors `OPENCODE_CONFIG_DIR` for the same reason). `configDir`, `configFile`, and `slashCommandPath` in the `claude-code` registry entry all derive from the resolved dir, so detection, install, refresh, and uninstall follow the same instance with no other call-site changes.

Wiring more than one instance is just one run per instance:

```bash
petdex hooks install                                    # default ~/.claude
CLAUDE_CONFIG_DIR=~/.claude-work petdex hooks install   # isolated work install
```

All instances then drive the same mascot through the shared sidecar, which already merges hook traffic from multiple agents.

## Verification

- `bun test` in `packages/petdex-cli`: 180 pass, including two new resolution tests mirroring the OpenCode ones
- `bun run build && bun run typecheck` in `packages/petdex-cli`: clean
- `bun run check` at root: clean
- Manual: importing `AGENTS` with `CLAUDE_CONFIG_DIR=/tmp/claude-b` yields `/tmp/claude-b/settings.json` + `/tmp/claude-b/commands/petdex.md`; unset falls back to `~/.claude`

### DIFF
--- a/packages/petdex-cli/src/hooks/agents.test.ts
+++ b/packages/petdex-cli/src/hooks/agents.test.ts
@@ -9,6 +9,7 @@ import {
   antigravityMcpConfigPath,
   applyCodexHooksFix,
   inspectFeaturesCodexHooks,
+  resolveClaudeConfigDir,
   resolveOpenCodeConfigDir,
   SIDECAR_URL,
 } from "./agents";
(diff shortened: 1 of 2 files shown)


Railly:
Thanks @KoalaBear, and good catch.

The CLI dropped its hook installer in v1: the desktop app wires agents
from its own Settings window now, so `hooks/agents.ts` no longer exists
for this to land on.

The bug does though. I checked, and the app hardcodes `$HOME/.claude` the
same way. Opened #601 for it with credit to you.

https://petdex.dev/download

Issue #569, opened 2026-07-17
petdex-desktop leaks zombie processes until the whole machine can't fork (sidecar respawn loop never reaps children)


zeapsu:
## Summary

`petdex-desktop` (macOS, v0.2.2) leaks one zombie (`<defunct>`) process every ~15–30s when the sidecar can't hold port 7777. Left running for a day, it exhausts `kern.maxprocperuid` (2666 on stock macOS) and **every fork on the machine starts failing** with `fork failed: resource temporarily unavailable`.

Observed on macOS 25.5.0 (arm64): after ~17h the desktop process had accumulated **2083 zombie children** (2659/2666 process slots used):

```
$ ps -u $USER -o comm= | sort | uniq -c | sort -rn | head -1
2083 <defunct>
$ ps -u $USER -o ppid=,stat= | awk '$2~/Z/{print $1}' | sort | uniq -c
2080 68886        # 68886 = petdex-desktop
```

## Root cause

`spawnSidecar` discards the child handle and never waits on it (`packages/petdex-desktop/src/main.zig:1430`):

```zig
_ = std.process.spawn(io, .{ .argv = argv, ... }) catch |err| { ... };
// Detach: we never wait on the child explicitly.
```

On POSIX, a child whose parent neither `waitpid()`s nor ignores `SIGCHLD` stays a zombie until the parent exits. That's harmless for the *one* healthy sidecar — the leak comes from the respawn loop:

1. WebView probes `/health` every 5s; after 3 consecutive failures it invokes `petdex.respawn_sidecar` (`src/main.zig:962–989`).
2. `respawnSidecarCmd` → `spawnSidecar` → detached spawn, never reaped (`src/main.zig:1608–1430`).
3. If the new sidecar can't bind :7777 — e.g. an orphaned incumbent still holds the port — `server.js` intentionally `process.exit(1)`s ("incumbent alive, yielding"). The dead node becomes a zombie, `/health` keeps failing, and the loop respawns forever: **~2 zombies/min, unbounded**.

The 4s respawn backoff doesn't help — the fail cycle itself is 15s, and each iteration leaks a process slot.

## Impact

System-wide DoS of the user account: once the per-user process table fills, *unrelated* apps can't fork (shells can't run commands, etc.). Killing petdex-desktop reparents the zombies to launchd, which reaps them instantly.

## Suggested fix

- Keep the sidecar `Child` handle (e.g. on `PetdexState`) instead of `_ =` discarding it.
[...]

PR #593, opened 2026-07-23
feat(desktop): opt-in chime when the pet starts waiting for input

KoalaBear:
> **Stacked on #591** — the first commit here is that PR (both touch the settings window and `writeDesktopSettingsCmd`, so basing on main would just manufacture conflicts). Only the last commit is new; happy to rebase onto main the moment #591 lands, or fold both if you'd rather review them together.

## Problem

The `waiting` state (permission prompts, idle alerts — the agent is blocked on you) is easy to miss: the pet is small and usually behind the window you're actually working in, so a prompt can sit unanswered for minutes while the agent idles.

## Fix

An opt-in chime for exactly that moment. `waitingSound` preference (default **false**) in `~/.petdex/preferences.json`, surfaced as a "Waiting sound" toggle in the desktop settings window next to the other toggles. Off by default on purpose — a desk pet that starts beeping unprompted after an update is a jump scare, not a feature.

Mechanics:

- The chime fires in the sidecar's `writeState` — the single point where a state is actually applied (queue drain, timed reset, boot) — and only on the **edge** into `waiting`. Re-posted waiting events for the same unanswered prompt stay silent; once you respond the state leaves `waiting`, so the next prompt is a fresh edge and chimes again. No timers, no debounce state to tune.
- Playback is a detached `afplay` of a stock macOS system sound (`Glass.aiff`): no dependencies, no bundled asset, and the sidecar already assumes macOS everywhere else (hdiutil, .app bundles). Non-darwin platforms no-op. The child is reaped by Node's normal SIGCHLD handling, so this can't reproduce the #569 zombie pattern.
- The preferences file is only read when a waiting edge actually happens — never on the per-tool-call state churn.
- `readDesktopBoolPreference` (from #591) now takes a per-key default, since this is the first preference that defaults to false.

## Verification

- `bun test` at repo root: 474 pass (new `waiting-chime.test.ts` covers the edge detection, the afplay invocation via an injected spawner, the non-darwin no-op, and the throwing-spawner guard; `update-utils` tests extended for the new preference)
- `cd packages/petdex-desktop && zig build && zig build test`: compiles, 18/18 pass (zig 0.16.0, zero-native sibling checkout per README)
- `bun run check`: clean
- `server.js` rebuilt with the sidecar's `bun run build`


Railly:
Thanks @KoalaBear, and sorry for the timing on this one.

The desktop rewrite removed `packages/petdex-desktop`, so this cannot merge
as-is. The idea is still good and nothing like it exists in the new app
yet, so it is worth reopening against `packages/petdex-desktop-native` if
you feel like porting it.

New app: https://petdex.dev/download

PR #598, opened 2026-07-25
fix(cli): make Claude Code hooks work on Windows via exec form

Subhajit936:
## Problem

On Windows, `petdex init` / `petdex hooks install` wires up Claude Code but the hooks don't work.

`bubbleHookCommand()` (in `src/hooks/agents.ts`) emits a **POSIX shell** snippet that gets written verbatim into `~/.claude/settings.json`:

```sh
[ -f "$HOME/.petdex/runtime/hooks-disabled" ] && exit 0; if [ -x "$HOME/.petdex/bin/petdex.js" ]...; then node ...; else curl ...; fi
```

Per the [Claude Code hooks docs](https://code.claude.com/docs/en/hooks.md), on Windows a `"type": "command"` hook runs through **Git Bash** when Git for Windows is installed, but falls back to **PowerShell** otherwise. `[ -f ]`, `if [ -x ]; then…fi`, and `$HOME` aren't portable across both, so hooks error out.

## Fix

Add `bubbleHookAction()`. On `win32` it emits Claude Code's **exec form** instead of a shell string:

```jsonc
{ "type": "command", "command": "node",
  "args": ["C:\Users\<you>\.petdex\bin\petdex.js", "bubble", "pre", "claude-code"] }
```

Exec form runs `node` directly with **no shell involved**, so there's nothing to be POSIX/PowerShell-incompatible. This is safe because:

- **stdin is still piped** → `runBubble` drains the hook payload exactly as before.
- **the killswitch still works** → `runBubble` already checks `hooks-disabled` first (`bubble-runner.ts`), so the shell-level check was redundant.
- **the path is native** → `path.join(homedir(), ...)`.

POSIX platforms keep the existing shell snippet unchanged. Scoped to the `claude-code` agent only; other agents are untouched.

The one behavior dropped on Windows is the raw-`curl` fallback for a missing persisted binary — which can't happen immediately after `init` persists it.

## Also: non-interactive install

Added a `--agents <id[,id]>` flag to `petdex hooks install`:

```bash
petdex hooks install --agents claude-code
```

The interactive `@clack` multiselect blocks on EOF in a non-TTY (Windows terminals, CI), so there was no way to wire hooks headlessly. The flag skips the prompt; the interactive path is unchanged when the flag is absent.

## Testing

- `bun run build && bun run typecheck` — clean
- `bunx biome check` on changed files — clean
- Installed on Windows 11 (Claude Code): desktop `win32-x64` binary launches, `hooks install --agents claude-code` writes exec-form hooks, and firing each hook returns exit 0. Verified the pet state transitions end-to-end (`jumping` → `waving`) with the sidecar logging `source=claude-code`.

### DIFF
--- a/packages/petdex-cli/bin/petdex.ts
+++ b/packages/petdex-cli/bin/petdex.ts
@@ -1411,7 +1411,30 @@ async function cmdHooks(args: string[]) {
   }
   switch (sub) {
     case "install": {
-      const { installedAgents } = await runHooksInstall();
+      // Non-interactive selection: `--agents claude-code,codex` or
+      // repeated `--agent <id>`. Lets Windows/CI wire hooks without a
+      // TTY (the interactive multiselect blocks on EOF there).
+      const agentFlag = args.find((a) => a.startsWith("--agents="));
+      let agents: string[] | undefined;
+      if (agentFlag) {
+        agents = agentFlag
+          .slice("--agents=".length)
+          .split(",")
+          .map((s) => s.trim())
+          .filter(Boolean);
+      } else {
+        const idx = args.indexOf("--agents");
+        const idx2 = idx === -1 ? args.indexOf("--agent") : idx;
+        if (idx2 !== -1 && args[idx2 + 1]) {
+          agents = args[idx2 + 1]
+            .split(",")
+            .map((s) => s.trim())
+            .filter(Boolean);
+        }
+      }
+      const { installedAgents } = await runHooksInstall(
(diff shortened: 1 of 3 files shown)


Railly:
Thanks for digging into the Windows path, @Subhajit936.

The CLI no longer installs hooks (v1 dropped that), and Windows is now a
first-class target for the desktop app itself, which writes them. So this
fix does not have a place to land anymore.

Windows build: https://petdex.dev/download

PR #602, opened 2026-07-26
fix(desktop): respect `CLAUDE_CONFIG_DIR` when wiring Claude Code hooks

### LINKED ISSUE #601: Desktop app ignores CLAUDE_CONFIG_DIR when wiring Claude Code hooks
`agent_hooks.zig:113` resolves Claude Code's config as `$HOME/.claude`
unconditionally, so anyone who moved it with `CLAUDE_CONFIG_DIR` gets
hooks written to a directory Claude Code does not read. Detection has
the same blind spot: that user shows as disconnected even after
connecting.

Found by @KoalaBear, who fixed it in the CLI in #592. That PR cannot
land because the CLI no longer installs hooks, but the bug moved into
the app with the feature.

Fix is the same shape: read `CLAUDE_CONFIG_DIR` and fall back to
`$HOME/.claude`.

dreyfus92:
the app always wrote Claude Code hooks to `~/.claude`, so if you run Claude Code from a custom `CLAUDE_CONFIG_DIR`, the hooks landed in a dir it never reads and the app kept showing you as disconnected. Now it reads `CLAUDE_CONFIG_DIR` first and falls back to `~/.claude` covered by a new test in `agent_hooks.zig` ✌🏻

one caveat that's worth noting 
>[!NOTE]
>apps launched from Finder don't see shell exports, so use `launchctl setenv` for that case or we can add a `settings-file` override later if anyone actually needs it.

closes: #601

### DIFF
--- a/packages/petdex-desktop-native/src/agent_hooks.zig
+++ b/packages/petdex-desktop-native/src/agent_hooks.zig
@@ -55,6 +55,18 @@ pub const AgentInfo = struct {
 
 pub const agent_count = 4;
 
+/// Set once in main() from CLAUDE_CONFIG_DIR (Zig 0.16 has no global
+/// getenv; env rides std.process.Init). Empty/unset means `~/.claude`.
+/// Claude Code supports fully isolated installs through this var, so
+/// writing to the default dir would wire hooks into a config the
+/// user's actual instance never reads.
+pub var claude_config_dir: ?[]const u8 = null;
+
+fn claudeDir(buf: []u8, home: []const u8) ?[]const u8 {
+    if (claude_config_dir) |d| if (d.len > 0) return d;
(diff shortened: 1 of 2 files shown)


Railly:
Thanks for this. Closing in favor of #607, which reaches the same behavior by a different route.

I ran both locally with `zig test src/agent_hooks.zig`: 6/6 pass on each. I then wrote three cases neither PR covered (override alongside a real `~/.claude`, override pointing at a directory that does not exist, blank override) and both behave identically on all three. The choice was not about correctness.

I went with #607 because its test covers unset and blank explicitly, and because it surfaced that `native test` does not collect the `agent_hooks.zig` tests, which is a harness gap we needed to know about.

Your Finder and `launchctl setenv` caveat is real and not covered by #607. Tracked in #615.

PR #607, opened 2026-07-28
fix(desktop): honor CLAUDE_CONFIG_DIR when wiring Claude Code hooks

### LINKED ISSUE #601: Desktop app ignores CLAUDE_CONFIG_DIR when wiring Claude Code hooks

KoalaBear:
Redo of #592 against the native app, per #601 — thanks @Railly for carrying the report over (and for the credit).

## Problem

`agent_hooks.zig` resolves Claude Code's config as `$HOME/.claude` unconditionally. Claude Code supports fully isolated side-by-side installs through `CLAUDE_CONFIG_DIR` — separate settings, separate slash commands, even separate logged-in accounts — so for anyone using that, Connect writes a `settings.json` their Claude Code never reads, and the Agents section shows them as disconnected even after a successful connect. Same blind spot in all three paths: detection (`scan`), install, and uninstall.

## Fix

Same shape as the CLI fix in #592: read `CLAUDE_CONFIG_DIR`, fall back to `$HOME/.claude`. The env var is snapshotted once in `main()` from `init.environ_map` (the `env_home` pattern — Zig 0.16 has no global getenv) and `scan`, `installClaude`, and `uninstall` all resolve through the same two helpers, so detection can never disagree with install about which instance they are talking about. Unset and set-but-blank both fall back.

## Tests

New test drives the whole loop against a fixture home with **no** `~/.claude` at all — the exact machine #601 describes: install lands in the override dir (and nothing leaks into `~/.claude`), scan reports `current` there, uninstall clears the override config, and null/blank both fall back to `~/.claude`.

One heads-up: `native test` only collects tests from `main.zig` (the root module file), so the `agent_hooks.zig` tests — the pre-existing ones included — don't run under it. Verified with `zig test src/agent_hooks.zig` directly: 6/6 pass. Happy to wire up a root-level test aggregator in a follow-up if you want them in the harness.

`native build` and `native test` both green against the pinned SDK (65d4885).

Fixes #601.

### DIFF
--- a/packages/petdex-desktop-native/src/agent_hooks.zig
+++ b/packages/petdex-desktop-native/src/agent_hooks.zig
@@ -55,6 +55,30 @@ pub const AgentInfo = struct {
 
 pub const agent_count = 4;
 
+/// Claude Code keeps everything under ~/.claude unless CLAUDE_CONFIG_DIR
+/// points elsewhere — that env var is how people run several fully
+/// isolated Claude Code installs (separate settings, separate accounts)
+/// on one machine. Resolving it here means detection, install, refresh,
+/// and uninstall all target the instance the user actually launches
+/// instead of always the default one (#601). Snapshot set once from
+/// main()'s environ_map, the same pattern as env_home: Zig 0.16 has no
+/// global getenv.
+pub var env_claude_config_dir: ?[]const u8 = null;
+
+fn claudeConfigDir(buf: []u8, home: []const u8) ?[]const u8 {
+    if (env_claude_config_dir) |dir| {
+        if (dir.len != 0) return std.fmt.bufPrint(buf, "{s}", .{dir}) catch null;
+    }
+    return std.fmt.bufPrint(buf, "{s}/.claude", .{home}) catch null;
+}
+
+fn claudeSettingsPath(buf: []u8, home: []const u8) ?[]const u8 {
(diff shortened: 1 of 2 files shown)


Railly:
Merging. Verified with `zig test src/agent_hooks.zig` (6/6) plus three extra cases: override alongside an existing `~/.claude`, override pointing at a directory that does not exist, and blank override. All three behave as you describe, and nothing leaks into the default config.

Your harness finding is confirmed: `native test` does not collect the `agent_hooks.zig` tests, so these never ran in CI. Worse, the repo has no Zig CI at all. The only check is the Vercel deploy, and it fails on authorization across every open PR. I am adding the workflow. The root-level test aggregator you offered would sit well on top of it.

On #602: closed in favor of this one. Credit to @dreyfus92, who landed the same behavior in parallel.

PR #608, opened 2026-07-28
feat(desktop): opt-in chime when the pet starts waiting for input

KoalaBear:
Port of #593 to the native app, per your invite there @Railly. The bubble-toggle half of that stack shipped with the rewrite, so this is just the chime, rebuilt on `applyState`.

## Problem

The `waiting` state (permission prompts, idle alerts — the agent is blocked on you) is the one state that's about *you*, and it's easy to miss when the pet sits on a second monitor or behind the editor. The whole point of the pet is glanceability; `waiting` deserves to reach you even when you aren't glancing.

## Fix

A "Waiting sound" toggle in Appearance, right under the activity-bubble one. Plays the platform's stock event sound when the pet *enters* `waiting`.

- **Off by default.** Sound is intrusive in a way passive UI never is — a desk pet that starts beeping unprompted after an update is a jump scare, not a feature. Persisted as `waiting_sound` in `desktop-native-settings.json`; a missing key stays silent.
- **Edge, not level.** `waiting` re-posts while one prompt sits unanswered (one Notification hook each), and a ping per re-post would train users to turn the feature straight back off. The dwell path in `poll_tick` refreshes an already-waiting state without touching `applyState`, so the edge check there sees real transitions only — once you respond, the state leaves `waiting`, and the next prompt is a fresh edge that chimes again. The predicate is a pure `shouldChime` with a test (in `main.zig`, so `native test` actually collects it).
- **No new dependency, no zombie.** Playback rides `fx.spawn` — `afplay` + Glass.aiff on macOS, `canberra-gtk-play --id complete` on Linux (silent no-op where canberra isn't installed), `rundll32 user32.dll,MessageBeep` on Windows. The exit Msg means the child is always reaped, so this can't reproduce the Zig-side zombie accumulation from #569, and a machine that can't play a sound just stays quiet.
- **Preview on enable.** Flipping the toggle on plays the chime once — the only way to judge a sound is to hear it, and hunting for a permission prompt just to preview a volume level is busywork.

## Verified

Beyond the unit test: built with `-Dautomation`, ran the app under a scratch `$HOME` with `waiting_sound: true`, cycled states via `native automate native-command petdex.cycle` — `afplay /System/Library/Sounds/Glass.aiff` spawned exactly on the transition into `waiting` and not on any other. `native build` and `native test` green against the pinned SDK (65d4885).

### DIFF
--- a/packages/petdex-desktop-native/src/main.zig
+++ b/packages/petdex-desktop-native/src/main.zig
@@ -131,6 +131,8 @@ pub const Msg = union(enum) {
     open_pet_page: u32,
     appearance: native_sdk.platform.Appearance,
     toggle_bubbles,
+    toggle_waiting_sound,
+    chime_done: native_sdk.EffectExit,
     install_agent: u32,
     uninstall_agent: u32,
     pet_filter: canvas.TextInputEvent,
@@ -141,7 +143,7 @@ pub const Msg = union(enum) {
     dismiss_install_error,
     noop,
 
-    pub const view_unbound = .{ "frame_tick", "poll_tick", "physics_tick", "frame_clock", "cycle_state" };
+    pub const view_unbound = .{ "frame_tick", "poll_tick", "physics_tick", "frame_clock", "cycle_state", "chime_done" };
 };
 
 pub const Model = struct {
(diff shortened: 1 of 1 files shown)


KoalaBear:
Pushed one addition: a single follow-up chime when a prompt has sat unanswered for two minutes (same waitingSound opt-in, once per waiting spell, resets when the user responds). The first ping is easy to miss mid-flow; a still-up prompt two minutes later really was missed. Predicate is pure and tested next to shouldChime.

Railly:
Merging. Off by default is the right call, and the edge-not-level reasoning holds up.

Verified on top of the post-#607 main: merges clean, and `zig test src/main.zig` passes both tests (chime fires only on the transition into `waiting`, escalation pings once).

Heads up: this will conflict with #610, #611 and #612 in `main.zig`. All four add `Msg` variants and `Model` fields on the same lines. The conflicts are mechanical, not logical: keep both sides. I confirmed this by running the merges in a worktree.

Opened #613 to split `main.zig`, which is the underlying cause.

PR #610, opened 2026-07-28
feat(desktop): bubble text size slider in Appearance

KoalaBear:
## Problem

The pet scales, the bubble text does not: it renders at a fixed 13pt with no way to raise it. On a large display (or for anyone who just wants bigger type) the activity bubble is squint territory.

## Fix

A "Bubble text size" slider in Appearance, right under Pet size: a real 13..20pt range, persisted as `bubble_text` in `desktop-native-settings.json`. 13 stays the default, so existing installs render pixel-identically until the slider moves.

**How, given the token system:** per-widget sizes only step ±1pt around the body base (13/14/15), and scaling the body base itself would drag the whole settings window along. The `.heading` typography rung is unused anywhere in this app — section titles sit on `.lg` — so `petdexTokens` aims it at the persisted preference and the bubble renders at `.size = .heading`. The bubble gets a real range; every other window keeps stock type. If the settings window ever wants a true heading rung this needs a revisit, so it's called out loudly in a comment at the assignment.

The 26-char wrap budget was tuned for 13pt over the fixed card width, so it now scales inversely with the glyph size (bigger type, fewer columns) — `bubbleCharsPerLine` is pure and tested, and the test lives in `main.zig` where `native test` actually collects it. The fixed 340x150 bubble window carries four honest lines at 20pt with room to spare, so no window/tail geometry moves.

## Verified

`native build` + `native test` green against the pinned SDK (65d4885). Ran the app under a scratch `$HOME` with `bubble_text: 17` and screenshotted the settings canvas via `native automate`: slider renders under Pet size with the thumb at the right fraction.

### DIFF
--- a/packages/petdex-desktop-native/src/main.zig
+++ b/packages/petdex-desktop-native/src/main.zig
@@ -133,6 +133,7 @@ pub const Msg = union(enum) {
     toggle_bubbles,
     toggle_waiting_sound,
     chime_done: native_sdk.EffectExit,
+    set_bubble_text_size: f32,
     install_agent: u32,
     uninstall_agent: u32,
     pet_filter: canvas.TextInputEvent,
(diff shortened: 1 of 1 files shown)


Railly:
Rebased onto main for you (maintainer edits were on), so this was not left waiting on a conflict it did not cause. Your commit and authorship are unchanged, still +67/-9 in one file.

The conflict was #608 and #616 landing in the same regions. Most hunks were keep-both, but one needed real merging rather than either side: `saveSettings` writes a single JSON line, and both PRs appended a field to it. It now carries both `waiting_sound` and `bubble_text`, and I checked the read path too, so the round trip is closed in both directions.

Verified after the rebase: `zig test src/main.zig` runs 23/23 (the 22 on main plus your `bubbleCharsPerLine` test), and `native build` is green.

One note for later, not a blocker: aiming `.heading` at the bubble is a clever use of an unused rung, and your comment says so at the assignment. If the settings window ever wants a real heading, that is the line to revisit.

Merging.

PR #611, opened 2026-07-28
feat(desktop): menu-bar residence — tray item and a Hide Dock icon toggle

KoalaBear:
## Problem

Petdex is ambient by nature, but it occupies a Dock slot and an app-switcher entry like a document app. And once the pet window is closed the app keeps running (by design — the hook server lives) with **no visible handle at all**: nothing in the menu bar, nothing to reopen Settings from, quit only by force.

## Fix

Two pieces that make Petdex behave like the menu-bar-resident tool it already half is:

**A status item, installed unconditionally** — Open Settings / Quit Petdex. The icon is the brand mark's silhouette with the pet face punched out (`src/assets/tray-icon.svg`, evenodd knockout of `public/brand/petdex-mark.svg`'s shapes): the host loads tray icons with `template = YES`, so only alpha survives — the gradient body would flatten into a blob otherwise. The platform takes a *path*, not bytes, and a relative path dies the moment Finder launches the bundle with `cwd=/`, so the embedded PNG is materialized into `~/.petdex/runtime/tray-icon.png` at boot and the tray points at the absolute path. Failure degrades to the host's title fallback, never a broken item.

**"Hide Dock icon" in Appearance (macOS only)** — flips the activation policy to `.accessory`: Petdex leaves the Dock and app switcher and lives in the menu bar, windows keep working. Off by default, persisted as `hide_dock`, applied at boot. The SDK host pins the policy to Regular and exposes no channel for it, so `plat.zig` reaches AppKit through the objc runtime, dispatched onto the main queue (NSApplication is main-thread-only; callers sit on the runtime loop thread). If you'd rather grow this into the SDK as a proper host API I'm happy to redo it on top — this is the app-side version that works on the current pin.

The ordering is deliberate: the tray exists *so that* the toggle can never strand anyone. Settings and Quit stay reachable with the Dock icon hidden and the pet window closed — which also finally gives "close the pet, keep the hooks running" a visible anchor.

One quit-path find worth flagging: SIGTERM rides the host's graceful-shutdown dispatch source, but that source is kqueue-backed and `EVFILT_SIGNAL` only sees **process-directed** signals — an in-process `raise()` is `pthread_kill(self)` and gets silently dropped. Verified both ways (external `kill -TERM` quit cleanly, `raise()` did nothing); `requestQuit` uses `kill(getpid(), SIGTERM)`.

## Verified

Against the pinned SDK (65d4885), scratch `$HOME`:
- tray icon materialized, status item installs warning-free
- toggling the switch persists `hide_dock:true` and the running app drops from LaunchServices' Foreground list
- booting with `hide_dock:true` registers straight as `type="UIElement"` (`lsappinfo`)
- `petdex.quit` exits cleanly through the graceful path (exit 0)
- `native build` + `native test` green; Appearance row is comptime-gated to macOS so Linux/Windows builds carry no dead toggle


Railly:
Rebased onto main for you (maintainer edits were on). Authorship and diff unchanged: +152/-3 across the same 5 files.

Three of the seven conflicts needed real merging rather than keeping both sides, so worth naming in case you review the result:

- `saveSettings` writes one JSON line and three PRs have now appended a field to it. It carries `waiting_sound`, `bubble_text` and `hide_dock`, and I verified all three read back.
- `view_unbound` is a tuple, not a list of lines: it is now the union, `chime_done` and `quit_app`.
- Your `hide_dock` load check and #608's `waiting_sound` check shared a closing brace. Keeping both verbatim would have left an unclosed `if` that still parsed as something else. They are two separate blocks now.

Verified: 23/23 tests, `native build` green.

On the objc seam for the activation policy: I would take you up on growing it into the SDK as a proper host API. Not a blocker for this, but `plat.zig` reaching into AppKit directly is the kind of thing that breaks quietly on an SDK bump.

The tray-before-toggle ordering is the right call. Merging.

PR #612, opened 2026-07-28
feat(desktop): quality of life — position memory, launch at login, focus mode

KoalaBear:
> **Stacked on #611** — the tray this adds Focus Mode to arrives there (both touch the status item and the objc seam in plat.zig, so basing on main would just manufacture conflicts). Only the last commit is new; happy to rebase onto main the moment #611 lands, or fold them if you'd rather review together.

Three small comforts:

## Position memory

The pet boots center-screen every launch and forgets where you put it. The window origin now persists as `pet_x`/`pet_y`, saved when a gesture *settles* (drag drop, throw rest — never per physics frame), restored once right after the first window fit (the fit re-anchors `bottom_center`, so an earlier move would be anchored away). The keys are only written once the window has actually been fitted and read, so a first-frame save can never pin `(0,0)`; missing keys keep today's default placement, and off-screen values from an unplugged monitor are left to the platform's own edge clamping — the same handling the throw physics already rides.

## Launch at login (macOS)

A settings toggle over `SMAppService` (macOS 13+; the class simply doesn't resolve below, and the row no-ops). ServiceManagement isn't among the frameworks the SDK links, so `plat.zig` dlopens it and reaches the API through the objc runtime — the `setDockIconHidden` route from #611. The model mirrors the **service's status**, never a stored wish: boot and every toggle re-query, so a rejected registration (user-declined approval in System Settings) snaps the switch visibly back instead of lying.

## Focus mode

One tray click mutes the bubble while the sprite keeps reacting — screen shares, recordings, busy sessions — no settings window round-trip. Session-only by design: focus ends when the app restarts. The tray menu is now model-derived (`status_item_fn`, static options keep icon+tooltip per the documented split) so the item shows On/Off. If #608 lands too, the waiting chime should gate on it as well — happy to wire that in whichever lands second.

## Verified

Live against the pinned SDK (65d4885), scratch `$HOME`:
- booted with `pet_x:600,pet_y:500`, then flipped an unrelated toggle: re-saved file still carries exactly `600/500` — the boot restore landed on the real window, round-trip closed
- Launch at login: toggled on → `SMAppService` registered and the switch read back `on` from the status query; toggled off → unregistered, switch back off (both directions on a real login-items DB)
- Focus Mode toggled twice via its command, app stable; quit path clean
- `native build` + `native test` green; both new rows comptime-gated to macOS

### DIFF
--- a/packages/petdex-desktop-native/src/main.zig
+++ b/packages/petdex-desktop-native/src/main.zig
@@ -135,6 +135,8 @@ pub const Msg = union(enum) {
     chime_done: native_sdk.EffectExit,
     set_bubble_text_size: f32,
     toggle_hide_dock,
+    toggle_launch_at_login,
+    toggle_focus_mode,
     quit_app,
     install_agent: u32,
     uninstall_agent: u32,
@@ -146,7 +148,7 @@ pub const Msg = union(enum) {
     dismiss_install_error,
     noop,
 
-    pub const view_unbound = .{ "frame_tick", "poll_tick", "physics_tick", "frame_clock", "cycle_state", "chime_done", "quit_app" };
+    pub const view_unbound = .{ "frame_tick", "poll_tick", "physics_tick", "frame_clock", "cycle_state", "chime_done", "quit_app", "toggle_focus_mode" };
 };
 
 pub const Model = struct {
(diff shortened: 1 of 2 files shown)


Railly:
Rebased onto main for you. #611 landed first, so the stacked commit dropped out and this is now the single QoL commit, +138/-11.

Two conflicts needed merging rather than either side:

- `view_unbound` is now the union of all three: `chime_done`, `quit_app`, `toggle_focus_mode`.
- `saveSettings` keeps your `pos` preamble and the conditional position keys, with `waiting_sound` and `bubble_text` from #608 and #610 folded into the same line. I checked the worst case against the 320-byte buffer: a 63-char slug with every field present and negative coordinates comes to 216 bytes, so there is 104 to spare. Worth knowing since a `bufPrint` overflow here fails silently and drops the whole save.

Verified: 23/23 tests, `native build` green.

Your note about gating the chime on Focus Mode: #608 is in, so that follow-up is now live work. Focus Mode being session-only is the right default.

Merging.

PR #614, opened 2026-07-28
fix(desktop): stabilize Linux Wayland transparency and overlays

ThermalEng:
## Summary

This PR stabilizes the native Petdex desktop client on Linux/KDE Wayland using the Linux runtime support now shipped by Native SDK pin `5f9f7cfa`.

Windows and macOS retain the behavior already present on `main`; all platform-specific application changes are gated to Linux.

## Linux Wayland behavior

- Use the Linux software presenter with premultiplied BGRA alpha.
- Keep the pet in a fixed transparent main toplevel.
- Present the existing dynamically sized companion bubble as a compositor-managed popup through `popup_parent = "main"`.
- Preserve #625's `bubbleWindowWidth(model)` / `bubbleWindowHeight(model)` sizing and configurable typography/layout.
- Leave Win/mac companion-window positioning and resizing unchanged.
- Keep Linux popup positioning parent-local instead of attempting unavailable Wayland global moves.

## Input and window integration

- Route Linux pet dragging through the Native SDK's compositor-owned window-drag command path.
- Map initial horizontal direction to `running-left` / `running-right`.
- Retain a 5-second watchdog for compositors that omit the release transition.
- Bind the existing context menu directly to the Linux pet hit route.
- Hide the GTK application menubar while keeping Win/mac menus unchanged.
- Use the existing status-item model with a full-color Linux tray icon.

## Rendering and logging

- Keep Linux background tokens fully transparent while leaving Win/mac theme backgrounds unchanged.
- Disable the unconditional Linux `on_frame` callback; material-state timers and the busy spinner still request redraws when needed.
- Build with `-Dtrace=off` in CI and release without changing the SDK default.
- Upload the Linux CI binary as an artifact for manual Wayland validation.

## SDK integration

- No SDK patch is carried in this PR.
- No `git apply` workflow step remains.
- Petdex `main` already pins Native SDK `5f9f7cfa` through #645, which contains Railly/native#1.

## Validation

- Rebases cleanly onto current upstream `main` (`8dd871e`).
- `native test`: 68/68 passed against SDK `5f9f7cfa`.
- Linux `native build -Dcpu=baseline -Dtrace=off`: passed (ReleaseFast).
- Windows cross-build `native build -Dtarget=x86_64-windows-gnu -Dcpu=baseline -Dtrace=off`: passed (ReleaseFast).
- Linux x86-64 binary installed and launched on KDE Wayland with no startup panic or SDK error.
- Installed binary SHA-256: `7b4199daa884a9c3e644c9a4ddaa08dddb7bfec1f0355166aaf802f5d41ba4b0`.
- `git diff --check`: passed.


Railly:
Reviewed the risky parts locally rather than by reading. Summary first: the work looks sound and I am not merging it yet, for one reason that is about blast radius rather than code quality.

## What I verified

- The patch applies cleanly to the pinned SDK (`65d4885`).
- The patched SDK's `zig build cli` succeeds on macOS.
- This PR's app builds green on macOS against that patched SDK.
- The SDK's own test suite fails identically with and without the patch, so those failures are pre-existing and not yours. I checked the unpatched pin as a control before saying so.

## The blocker

`git apply` is added to the SDK clone step with no platform condition, in **both** workflows including `desktop-release.yml`. The step has no `if:`, so every macOS and Windows build, and every tagged release, now compiles against a patched SDK too.

The patch is not Linux-only in reach. It touches 15 SDK files, and several are cross-platform: `primitives/canvas/widgets.zig`, `primitives/canvas/text_spans.zig`, `runtime/ui_app.zig`, `platform/types.zig`. macOS compiling is good news, but compiling is not the bar this repo uses, and the one thing nobody has run is a patched macOS build in front of a human.

## What would unblock it

Any of these, your call:

1. Gate the `git apply` on `matrix.platform == 'linux'` (and the release job on the Linux target). Smallest change, keeps macOS and Windows on the untouched pin.
2. Land the patch upstream in the SDK and move `NATIVE_SDK_REF`, which is what the pin comment in both workflows already asks for. Cleanest, slowest.
3. Keep it global, and post macOS and Windows screenshots from the patched build.

## Also worth splitting

The CJK font work (`NotoSansSC`, the OFL text, the bubble text scaling) is independently useful and has nothing to do with Wayland alpha. As its own PR it would land in a day.

## One process note

The Zig CI never ran on this PR: the only checks here are Vercel. Your branch is `ThermalEng/petdex#main`, and the workflow triggers on paths that are not matching from that ref. So nothing has compiled this on Linux or Windows in CI either, which is the platform the PR is about. Worth pushing to a named branch so the three legs run.

Thanks for the depth here, the KWin findings and the `-Dcpu=baseline` note in particular. That last one turned out to be a real user-facing crash on AMD, tracked and fixed separately in #621.

ThermalEng:
Thanks for doing the patched/unpatched macOS control and for identifying the workflow blast radius.

The work is now split into three review boundaries:

1. This PR (#614) is Linux-only. The SDK patch is gated by `matrix.platform == 'linux'` in CI and `matrix.target == 'linux-x64'` in release, so Windows and macOS use the untouched pinned SDK.
2. The Linux SDK implementation is available independently in Railly/native#1: https://github.com/Railly/native/pull/1
3. The platform-neutral bubble size/rows/columns and multiline-hook changes moved to #625: https://github.com/crafter-station/petdex/pull/625

The Noto Sans SC asset, OFL file, font registration, and per-span font work are deferred and are not present in any of these PRs. The global SDK trace-default change was also removed; Petdex requests `-Dtrace=off` explicitly.

Validation after the split:

- The Linux-only ReleaseFast build succeeds with the patched SDK, `-Dcpu=baseline`, and `-Dtrace=off`, and was installed and manually launched on KDE Wayland.
- Windows `x86_64-windows-gnu` ReleaseFast cross-build succeeds against a pristine checkout of the untouched `65d4885` SDK pin.
- The common #625 branch independently builds for Linux and cross-compiles for Windows against that same untouched SDK.

The GitHub workflow runs were created but currently show `action_required`; they were not path-skipped and appear to require maintainer approval for fork-authored workflow execution.

ThermalEng:
Now that #625 and #626 have landed, could you clarify which SDK path you would prefer for #614 before I rebase it onto the latest `main`?

1. Merge Railly/native#1 into the SDK line and update Petdex's pinned `NATIVE_SDK_REF`; or
2. Keep #614's current Linux-only patch application, gated to the Linux CI and release legs.

I am happy to follow either direction. I would prefer to settle this first so the upcoming #614 conflict resolution does not have to be redone if the SDK integration strategy changes.

Railly:
Railly/native#1 is merged. The SDK pin can move to `5f9f7cfa`, which means the 1681-line patch file and both `git apply` steps come out of this PR entirely. Verified main against the new pin already: 55/55 tests, macOS and `x86_64-windows -Dcpu=baseline` both green.

I tried to rebase this for you and stopped, because the conflicts are not mechanical this time. Your own #625 changed the ground underneath it.

`main.zig`, five hunks:

1. **Bubble sizing.** #625 replaced the fixed `bubble_window_w/h` constants with computed `bubbleWindowWidth(model)` / `bubbleWindowHeight(model)`. This branch still uses the constants in four places, including `fitWindow`.
2. **The bubble is a separate window now.** Main builds a second `bubble` window descriptor with its own canvas; this branch deletes that block and folds the height back into the main window.
3. `tokens.typography` — yours sets `font_id`, main aims the `.heading` rung at the persisted bubble text size.
4. `.fonts = app_fonts` versus `&app_fonts`.
5. A one-line divergence in the bubbles-enabled path.

Picking sides on 1 and 2 is choosing between two bubble architectures, and you wrote both. That is yours to make, not mine to guess.

The two workflow conflicts are trivial and mine: I added comments above `-Dcpu=baseline` in #621 explaining the AVX-512 crash, and your `-Dtrace=off` edit dropped them. Keep both.

So: move the pin to `5f9f7cfa`, delete the patch and its two apply steps, and rebase onto main resolving toward #625's bubble model. That should leave this as `main.zig` plus the workflow `-Dtrace=off` and the Linux artifact upload, which is a normal-sized PR.

Say the word if you would rather I take it and pick #625's side on both, and I will tell you exactly what I chose.

Railly:
Option 1, and the pin is moved: `NATIVE_SDK_REF` is `5f9f7cfa` on main as of #645.

Apologies for the delay. I told you on 08-02 that the pin could move and then did not move it, so you have been waiting five days on a one-line change from me. You were right to ask before rebasing: doing it against the old pin would have meant redoing the resolution afterwards.

So for the rebase: the patch file and both `git apply` steps come out entirely. The SDK carries that work now. What should be left is `main.zig` plus the workflow `-Dtrace=off` and the Linux artifact upload.

The `main.zig` conflicts are the ones I flagged earlier and they are still yours to call, since #625 is your own work: the bubble is a separate window sized by `bubbleWindowWidth(model)` now, not the fixed constants this branch still uses. If you would rather I take those and tell you exactly what I picked, say so and I will.

Verified main against the new pin before merging it: 68/68 tests, macOS and `x86_64-windows -Dcpu=baseline` green.

ThermalEng:
Rebased and resolved locally against current `main`, choosing #625/#626's separate dynamically sized bubble window as requested.

The SDK patch file and both `git apply` steps are gone; this now uses the `5f9f7cfa` pin already on main. The remaining diff is `main.zig`, the Linux tray icon, workflow `-Dtrace=off`, and the Linux test-binary artifact.

Validation completed before pushing: 68/68 tests, Linux ReleaseFast, Windows x86_64 cross-build, `git diff --check`, and a local KDE Wayland install/start with no startup error. The local binary was also built with `-Dtrace=off`. New head: `9bbeb84`.

The upstream checks triggered by this push are now green for lockfile/SDK-pin validation and macOS, Linux, and Windows builds.

ThermalEng:
Conflict-resolution details, matching the choices requested:

1. **Bubble sizing:** kept current main/#625's computed `bubbleWindowWidth(model)` and `bubbleWindowHeight(model)` model; dropped the legacy fixed-size and scale-derived Linux sizing helpers.
2. **Bubble architecture:** kept the separate companion bubble window. Linux adds only `popup_parent = "main"` and skips application-side global move/resize; Win/mac retain main's companion-window path.
3. **Typography:** kept both main's persisted `heading_size = model.bubble_text_px` and the custom `font_id`; added only the Linux transparent background token override.
4. **Fonts field:** kept current main/new-SDK form, `.fonts = app_fonts`.
5. **Bubbles-enabled path:** kept main/#626 behavior, including `clearBubble(model)` when messages are disabled.
6. **Workflow conflicts:** kept the upstream AVX-512/baseline comments, added `-Dtrace=off`, retained the Linux test-binary artifact, and removed the SDK patch plus both `git apply` steps.
7. **Newer main conflicts:** preserved the upstream first-pet install path and other post-#625/#626 additions, then added the Linux native-drag messages alongside them.

No source tests or temporary screenshot/bubble-injection test steps were added or retained.

Railly:
Superseded by #670, now merged. That is this branch rebased onto main with the three `main.zig` conflicts resolved. I opened it as a separate branch because this PR comes from your fork's `main`, and rebasing in place would have rewritten that branch for you.

Your commits are preserved and the merge carries `Co-authored-by`, so the credit stays with you.

How the conflicts were resolved, in each case keeping what landed after this branch was written:

- `gpu_alpha_mode` stays `premultiplied` on every platform. #662 changed it from `opaque` on purpose to remove the visible rectangle behind the transparent window, and locked it with a test. Your Linux `software` backend is preserved.
- `syncBubbleWindow` keeps the mutable model from #667, so the flip is decided where it is consumed, with your Linux popup early-return on top.
- The non-Linux bubble descriptor keeps `bubbleWantY` and `fullscreen_overlay` from #667 and #656. This branch still carried the pre-#667 geometry, which would have reverted both.

Worth knowing for future work: `build linux` failed on the first run of #670 because the test `transparent surfaces clear independently from settings` asserted an opaque settings background on every platform. That holds on AppKit and Win32, where the pet window owns its transparency and settings paints its own page background. Your Linux path clears every surface to zero alpha through GTK, so settings is transparent there too and the assertion failed on that runner only. It is now scoped per platform, and linux, macos and windows all pass.

Thanks for the Wayland work.

Issue #589, opened 2026-07-23
Add a CI-repeatable test for the PetSprite preview fallback


Railly:
Follow-up from #587 review.

The onError -> atlas swap shipped verified end-to-end by hand (headless Chrome DOM dump), but no test references `PetSprite`. A refactor of `src/components/pets/pet-sprite.tsx` could silently regress the fallback and cards would go blank again exactly like #579.

Suggested: a component test that renders PetSprite with a `fallbackSrc`, fires the probe img's `error` event, and asserts the sprite background swaps to the atlas URL and layout switches to `atlas`. Prove it red first by reverting the fallback logic (force-red rule).

PR #616, opened 2026-07-28
chore(desktop): actually run the native test suite — root aggregator + CI step

KoalaBear:
Found while working on #607: `native test` reported success while executing almost nothing.

## Problem

`zig build test` collects tests from the **root module file only**, and imports are analyzed lazily. main.zig has no aggregator, so every test in the imported files — agent_hooks' non-destructive settings.json merging, hook_runner's payload parsing, installer's URL allowlisting, **18 in all** — compiled green and never ran. Easy to prove: the agent_hooks fixture tests write `/tmp/petdex-agenthooks-fixture` as a side effect, and that directory exists on nobody's machine. On top, `desktop-native-ci.yml` builds and exercises the binary but has no test step, so even a working suite had nowhere to fail.

## Fix

- The standard root aggregator block at the bottom of main.zig (`test { _ = agent_hooks; ... }`) — referencing the imports forces their semantic analysis, which is what registers their tests. 19/19 pass locally against the pinned SDK, fixture side effects present.
- A `native test` CI step on the macOS and Linux legs. Windows sits out deliberately: the fixture tests write under `/tmp` by absolute POSIX path, and enabling that leg means teaching the fixtures a portable temp dir first — a separate, mechanical follow-up if you want it.

Relates to #589 (CI-repeatable tests) in spirit; touches only the test seam, zero behavior change.

### DIFF
--- a/.github/workflows/desktop-native-ci.yml
+++ b/.github/workflows/desktop-native-ci.yml
@@ -136,6 +136,22 @@ jobs:
           "$NATIVE_CLI" build
           ls -lh zig-out/bin/
 
+      - name: Run the test suite
+        # The suite ran on nobody's machine until the root aggregator
+        # existed: `zig build test` collects tests from the root module
+        # file only, so the imported files' tests (agent_hooks' config
+        # merging, hook_runner's payload parsing, installer's URL
+        # checks) compiled green and never executed — and CI had no
+        # test step to notice. Unix legs only for now: the fixture
+        # tests write under /tmp by absolute POSIX path, which Windows
+        # has no equivalent for until the fixtures learn a portable
+        # temp dir.
+        if: matrix.platform != 'windows'
+        working-directory: packages/petdex-desktop-native
+        shell: bash
+        run: |
+          "$NATIVE_CLI" test .
+
       # Compiling proves nothing about whether the app draws: the hook
       # server regression that shipped in this branch compiled fine on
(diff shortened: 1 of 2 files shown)


Railly:
Confirmed, and it is worse than the title suggests. I ran the suite on main before and after your aggregator:

- before: `All 0 tests passed.`
- after: `All 22 tests passed.`

Zero. Not 18 hidden among a few real ones: the whole suite was inert. That includes installer's path-traversal and asset-host allowlist tests, which are the ones you least want silently disabled.

This needs a rebase: #608 landed while you were writing this, and its two tests sit at the same end-of-file spot as the aggregator block. The conflict is one hunk and the resolution is to keep both sides. I verified that resolution locally, hence the 22 above.

Once this is green it goes in ahead of #610, #611 and #612, so those land with a suite that actually runs. Thanks for catching it while working on something else.

Railly:
I rebased this onto main for you (maintainer edits were on) so it was not blocked waiting on a one-hunk conflict. Your commit and authorship are unchanged, the diff is still your +30/-0, and the resolution is the one I described: keep #608's two tests, then your aggregator block.

Verified after the rebase: `zig test src/main.zig` runs 22/22, including the installer traversal and host-allowlist tests that had never executed.

Merging now so #610, #611, #612 and #617 land on a suite that actually runs.

Issue #557, opened 2026-07-15
✨ Feature Request: Interactive Pet Actions (Feed, Pet, Play)


jacobsonyang123:
## Summary

Currently, petdex pets are **state-driven animations** that react to agent activity (idle, run, review, error, done). This is great, but users often want to **interact** with their pets directly — feed them, pet them, play with them, etc.

## Motivation

- Users form emotional bonds with their desktop pets and want to engage beyond watching
- Interactive actions (like feeding a cat 🐱) would increase user engagement and delight
- Other virtual pet ecosystems (Tamagotchi, Neopets, desktop pets) have proven that interaction loops drive retention

## Proposed Interactions

| Action | Trigger | Example |
|--------|---------|---------|
| **Feed** |  or click | Cat eats fish, gains happy state |
| **Pet/Stroke** |  or click | Cat purrs animation |
| **Play** |  | Cat chases laser dot |
| **Sleep** |  | Cat curls up |

## Possible Implementation Ideas

1. **CLI commands**:  triggers a one-shot animation
2. **Desktop app**: Click on the pet sprite to trigger random interaction
3. **Mood system**: Track a simple mood/hunger stat that decays over time, affected by interactions
4. **Animation states**: Add new sprite states like , , ,  to the sprite sheets

## Context

I'm running the **big-orange-cat** 🐱 as my Hermes desktop pet and would love to feed it!

---

*Opened via Hermes Agent* 🤖

PR #617, opened 2026-07-28
feat(desktop): pat the pet — a plain click gets a reaction

KoalaBear:
First slice of #557 (Interactive Pet Actions).

## Problem

A left-click on the pet does nothing: the press is claimed by the drag gesture, and a release without movement falls through in silence. The most natural interaction with a desk pet — poking it — is dead air.

## Fix

Detect the tap on release: quick (≤400ms, the classic click ceiling) and the window never left the press point (3px slop for the device-pixel jitter a fast grab produces). Play a reaction, alternating waving/jumping so consecutive pats don't feel canned. The reaction rides `applyState`'s normal dwell and hands back to idle exactly like a throw-end wave.

Deliberately narrow:
- Drags and throws untouched — the tap check runs before the velocity branch and only claims releases that today do nothing.
- Right-click context menu untouched.
- No persistence, no economy. #557 asks for feed/play/stats too; those want pet-side art and probably a design pass, so this lands the interaction primitive first. Happy to iterate toward the rest if there's appetite.

The tap predicate is pure and tested (runs under `native test` — root file; #616 makes the rest of the suite run too).

## Verified

`native build` + `native test` green against the pinned SDK (65d4885). The gesture itself is cursor-polled, so it needs a human hand: click the pet → wave; click again → jump; drag still drags, throw still throws.

Closes nothing yet — first slice toward #557.

### DIFF
--- a/packages/petdex-desktop-native/src/main.zig
+++ b/packages/petdex-desktop-native/src/main.zig
@@ -184,6 +184,12 @@ pub const Model = struct {
     dragging: bool = false,
     grab_dx: f64 = 0,
     grab_dy: f64 = 0,
+    // Where and when the current grab began, for tap detection on
+    // release; pat_flip alternates the reaction between two states.
+    press_x: f64 = 0,
+    press_y: f64 = 0,
+    press_ms: i64 = 0,
+    pat_flip: bool = false,
     settings_open: bool = false,
     /// Sprite scale, persisted. Codex parity: the settings slider maps
     /// 0.4..1.2 over this.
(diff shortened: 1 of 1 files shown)


Railly:
Rebased onto main for you. One conflict, keep-both at the end of the file where every recent PR has been adding tests. Authorship and diff unchanged at +47/-0.

Verified: 24/24 tests including your `isTap` cases, `native build` green.

Taking you at your word that the gesture itself needs a human hand: the predicate is tested, the wiring is not, and I have not clicked a pet to confirm. That is an honest limit rather than a blocker, given the change only claims releases that currently do nothing.

Landing the interaction primitive before the feed/play economy is the right order. Merging.

PR #621, opened 2026-07-29
fix(desktop): build release binaries for baseline x86-64

Railly:
Fixes #604. Fixes #618.

## Problem

The `desktop-v0.3.1` Windows binary carries AVX-512. Disassembling the published exe (sha `2E734C70...`, the one @kongbaic checksummed) finds 862 `kmovd`, 199 `vpternlogd`, and 1512 AVX-512-only instructions in total. No Zen 3 part decodes those, which is why the app dies at startup with `0xC000001D` before it draws anything.

Two reports, two different AMD CPUs: @sasa-lty in #604 and @kongbaic on a Ryzen 5 5600G in #618.

Zig targets the host CPU unless told otherwise, and the GitHub runner is newer than the machines people install on. The runner's own CPU was leaking into the artifact.

## Fix

`-Dcpu=baseline` on the release build, and the same flag in CI so it exercises the instruction set users actually get.

## Verified

Three builds of the same source, disassembled and compared:

| build | AVX total | AVX-512 only | kmovd | vpternlogd |
| --- | --- | --- | --- | --- |
| shipped v0.3.1 | 67635 | 1512 | 862 | 199 |
| `-Dcpu=x86_64_v4` (control) | 67768 | 1517 | 862 | 199 |
| `-Dcpu=baseline` (this PR) | 0 | 0 | 0 | 0 |

The control reproduces the release almost instruction for instruction, which confirms the mechanism rather than assuming it. Baseline emits zero AVX of any kind.

## Note

I cannot run a Windows binary here, so this is proven at the instruction level, not by launching the app on a Ryzen box. Worth having @sasa-lty or @kongbaic confirm against the next release build.

#614 already applies the same flag for its Linux work. If that lands first, this becomes a no-op for the shared build step and only the CI leg differs.

PR #623, opened 2026-07-29
fix(desktop): get the pid without libc on Linux

### LINKED ISSUE #622: Linux CI red: native test's analysis step compiles the app without libc
Since #616 added the test step, the Linux leg of `desktop-native-ci` fails. Main is red, and every PR inherits it.

The tests themselves pass. The build summary says so on the same run:

```
Build Summary: 7/9 steps succeeded (1 failed); 22/22 tests passed
```

What fails is a sibling step:

```
std/c.zig:11013:12: error: dependency on libc must be explicitly specified in the build command
pub extern "c" fn getpid() pid_t;
referenced by:
    processId: src/plat.zig:267:31
    start: src/hook_server.zig:187:39
```

## Cause

`native test` runs a `<app>-analysis` object alongside the test binary to force semantic analysis of `main`. The SDK builds that module (`build/app.zig:609`) without `link_libc`, while the real app binary links it. `plat.processId` calls `std.c.getpid()`, which Linux requires an explicit libc dependency for and macOS does not. So the analysis object compiles on macOS and Windows and fails only on Linux.

`native build` passes on the Linux leg of the same run: the shipped binary is fine. This is a test-harness gap, not an app defect.

## Impact

The failure skips the steps that follow it: the hook-server round trip and the deep-link check on Linux. Those are the parts that verify behavior rather than compilation, so the leg that fails loudest is also the one that stops testing.

## Fix directions

- SDK side, the real fix: have the analysis module inherit the app's libc linkage.
- App side, if we want green before the SDK moves: give `processId` a Linux branch that does not go through libc.

Related to #613 in spirit: `plat.zig` is where the per-platform branches live.

Railly:
Fixes #622. Main has been red on the Linux leg since #616 landed.

## Problem

`native test` builds an `<app>-analysis` object alongside the test binary to force semantic analysis of `main`. That module does not link libc. `plat.processId` calls `std.c.getpid`, an `extern "c"` declaration that Linux refuses to compile without explicit linkage:

```
std/c.zig:11013:12: error: dependency on libc must be explicitly specified in the build command
pub extern "c" fn getpid() pid_t;
referenced by:
    processId: src/plat.zig:267:31
```

The app binary links libc, so `native build` passes on the same run. macOS and Windows do not require the explicit dependency, so only Linux failed. It stayed invisible until #616 added a test step that runs the analysis object.

The cost is larger than one red check: the failing step skips the ones behind it, which on the Linux leg are the hook-server round trip and the deep-link check. Those are the steps that verify behavior rather than compilation, so the platform that fails loudest is also the one that stops being tested.

## Fix

`std.os.linux.getpid` is the raw syscall. No linkage, same value. Windows and macOS keep their existing branches.

## Verified

Forced analysis of `processId` for `x86_64-linux` with no libc, which is what the analysis object does:

- before: reproduces the CI error above, same stdlib line
- after: compiles

Runtime checked as well, not just compilation: the pid the function returns matches the one the shell sees for the same process. `native test` and `native build` are green on macOS.

## Note

This is the app-side fix, and it unblocks CI today. The SDK-side fix is still worth doing: the analysis module should inherit the app's libc linkage, otherwise the next `std.c` call anywhere in the app reintroduces this. I left that direction in #622.

### DIFF
--- a/packages/petdex-desktop-native/src/plat.zig
+++ b/packages/petdex-desktop-native/src/plat.zig
@@ -261,9 +261,17 @@ pub fn replaceSymlink(target: []const u8, link: []const u8) bool {
 
 /// Own pid, for the /whoami endpoint. std has no portable accessor in
 /// 0.16, so this is the one genuine per-platform branch in this file.
+///
+/// Linux goes through the raw syscall rather than libc: `std.c.getpid`
+/// is an `extern "c"` declaration, and Linux refuses to compile one
+/// unless the build links libc explicitly. The app binary does, but
+/// `native test` also builds an analysis object that does not, so the
+/// libc path failed the Linux leg of CI while compiling fine on macOS
+/// and Windows. The syscall needs no linkage and returns the same pid.
 pub fn processId() u32 {
     return switch (builtin.os.tag) {
         .windows => std.os.windows.GetCurrentProcessId(),
+        .linux => @intCast(std.os.linux.getpid()),
         else => @intCast(std.c.getpid()),
     };
 }
--- a/packages/petdex-desktop-native/src/plat.zig
+++ b/packages/petdex-desktop-native/src/plat.zig
@@ -351,8 +351,6 @@ pub fn setLaunchAtLogin(enabled: bool) bool {
     return @as(MsgSendErr, @ptrCast(&AppleApp.objc_msgSend))(svc, sel, null);
 }
 
-extern "c" fn kill(pid: c_int, sig: c_int) c_int;
-
 /// Quit the way a menu quit would. SIGTERM rides the hosts'
 /// graceful-shutdown paths (the macOS host turns it into `NSApp
 /// terminate` through a dispatch source, sealing journals on the way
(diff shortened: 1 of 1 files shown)


Railly:
Scope grew: pushing the first commit turned CI red on **Windows**, and that one is not a harness gap. Main does not compile for Windows right now.

Both defects come from #611, which I merged. Neither could be caught at the time: the Linux leg was already red from #622, and the Windows leg never reached the analysis object.

- `kernel32.ExitProcess` is not declared in Zig 0.16's std. The Windows build fails outright. Now exits through `std.process.exit`, which is `RtlExitUserProcess` underneath.
- `requestQuit` declared its own `extern "c" fn kill`, the same libc dependency as `processId`. Linux now uses the raw syscall.
- Signals are an enum in 0.16, so `.TERM` rather than `15`. My first attempt used the integer and failed to compile, which is a good argument for the analysis object existing at all.

Verified per platform, forcing analysis with no libc linked plus real builds: Linux, macOS and Windows objects compile, `native build` is green for `x86_64-windows` and macOS, `native test` passes.

Worth stating plainly: this is a regression I let in. The Windows crash fix in #621 shipped into a tree that then stopped building for Windows, and only the test step added by #616 surfaced it.

Railly:
CI is green on all three legs, and this is the first fully green run of `desktop-native-ci` in this stretch.

The part that matters most is what the Linux leg now reaches. These steps had been skipped every run, because the failure sat in front of them:

```
success  Run the test suite
success  Install a pet (Unix)
success  Run and exercise the hook server (Linux)
success  Fire a real deep link (Linux)
```

Those are the steps that verify behavior rather than compilation. Linux has not actually been tested since #616 landed. Merging.

PR #624, opened 2026-07-29
feat(desktop): daily pet rotation and a tray Shuffle Pet

KoalaBear:
## Problem

People install several pets and then look at the same one forever. The variety lives in the catalog and never on screen — the app gives you no reason to ever see pet #2.

## Fix

Two doses of variety, both riding the existing `select_pet` Msg so activation cannot drift between a click, a rotation, and a shuffle:

**"Rotate pet daily"** (Appearance, off by default — existing installs change nothing): the pet advances round-robin through the installed catalog once per UTC epoch-day. Design calls, each deliberate:

- *Daily, not per-launch or per-interval:* an ambient companion changing identity mid-work is distraction, not variety; per-launch stopped meaning anything once launch-at-login landed (#612); daily is a small ritual — wake up to a different pet.
- *Epoch-day, not local calendar:* the flip lands at UTC midnight — night hours for most timezones — without dragging timezone math into the app. The check rides `poll_tick` against a persisted `rotation_day`, so an app closed (or asleep) over midnight still rotates on next run.
- *Round-robin, not random:* deterministic, no repeats until the catalog wraps, no RNG dependency.
- *Never on the spot:* enabling the toggle stamps the current pet as today's; rotation starts tomorrow. A manual pick stamps today too — choosing a pet pins it until the next day rather than fighting the rotation.

**"Shuffle Pet"** in the tray: instant advance, works independent of the daily toggle — variety on demand, one click.

`advancePet` skips pets whose sheets the codec refuses (`select_pet` only commits after a successful load) and gives up after one full loop, so a catalog of stubs cannot spin. The settings buffer grows to 384 with the worst case measured in a comment — noting your #612 observation that a `bufPrint` overflow there drops the save silently.

## Verified

Live against the pinned SDK (65d4885), scratch `$HOME`, two pets:
- booted with `rotate_pets:true, rotation_day:20600` (stale): first tick rotated doraemon → zebra and stamped `rotation_day:20663`
- `petdex.shuffle` twice: zebra → doraemon → zebra — round-robin wraps
- enabling the toggle does not swap the current pet
- 25/25 tests (epoch-day boundary test pinned to a real date so a unit slip cannot pass), `native build` green

### DIFF
--- a/packages/petdex-desktop-native/src/main.zig
+++ b/packages/petdex-desktop-native/src/main.zig
@@ -137,6 +137,8 @@ pub const Msg = union(enum) {
     toggle_hide_dock,
     toggle_launch_at_login,
     toggle_focus_mode,
+    toggle_rotate_pets,
+    shuffle_pet,
     quit_app,
     install_agent: u32,
     uninstall_agent: u32,
@@ -148,7 +150,7 @@ pub const Msg = union(enum) {
     dismiss_install_error,
     noop,
 
-    pub const view_unbound = .{ "frame_tick", "poll_tick", "physics_tick", "frame_clock", "cycle_state", "chime_done", "quit_app", "toggle_focus_mode" };
+    pub const view_unbound = .{ "frame_tick", "poll_tick", "physics_tick", "frame_clock", "cycle_state", "chime_done", "quit_app", "toggle_focus_mode", "shuffle_pet" };
 };
 
 pub const Model = struct {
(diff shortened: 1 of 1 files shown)


Railly:
First PR in this stretch that arrived with all three CI legs already green, which is what #616 and #623 were for. Merges clean onto main with no rebase needed.

Verified independently: 25/25 tests, `native build` green on macOS and cross-compiled for `x86_64-windows`. I also checked the buffer claim rather than taking it: worst case with a 63-char slug, every field present, negative coordinates and an 8-digit epoch-day comes to 261 bytes against your 384, so 123 to spare. Growing it was the right instinct.

The design calls hold up. Epoch-day over local calendar is the one I would have argued about, and your reason (the flip lands in night hours for most timezones without dragging tz math into the app) is better than the alternative. Stamping the current pet as today's when the toggle flips is the detail that makes it feel intentional rather than jumpy.

Merging.

PR #625, opened 2026-07-30
feat(desktop): make bubble layout configurable

ThermalEng:
## Summary

This PR makes native desktop bubble typography and layout explicitly configurable across Linux, Windows, and macOS.

## Changes

- Keep bubble text size as a manual preference and lower the minimum to 8 px.
- Add `Characters per line` (default 40, range 8–120).
- Add `Answer lines` (default 2, range 1–8); the title occupies one additional row.
- Persist `bubble_text`, `bubble_columns`, and `bubble_answer_lines`.
- Keep the visible card intrinsically sized to current content.
- Calculate the transparent companion-window upper bound from configured rows, columns, text size, padding, tail, icon slots, and safety margin.
- Split answers into explicit rows on UTF-8 boundaries so measured and painted line counts agree.
- Truncate the title without a redundant ellipsis beside the busy spinner; truncated answer text retains its final ellipsis.
- Preserve escaped newlines while scanning multiline agent hook JSON.
- Retain the animated busy spinner.
- Add an optional local TrueType font path, persisted as `font_path` in the existing settings JSON and applied to all app text after restart. Invalid files fall back to the default font.

## Scope and independence

- Based directly on upstream `main`.
- Does not include or depend on #614, #626, or Native SDK changes.
- Adds no Wayland, GTK, popup-parent, transparency, tray, renderer, or platform-specific window behavior.
- No bundled font is included.

## Validation

- Standalone Linux ReleaseFast build passed.
- Standalone Windows `x86_64-windows-gnu` ReleaseFast cross-build passed.
- The independent branch is mergeable against upstream `main`.

### DIFF
--- a/packages/petdex-desktop-native/src/hook_server.zig
+++ b/packages/petdex-desktop-native/src/hook_server.zig
@@ -133,8 +133,8 @@ pub const Mailbox = struct {
 pub var mailbox: Mailbox = .{};
 
 const valid_states = [_][]const u8{
-    "idle", "running", "running-left", "running-right", "waving",
-    "jumping", "failed", "review", "waiting",
+    "idle",    "running", "running-left", "running-right", "waving",
+    "jumping", "failed",  "review",       "waiting",
 };
 
 /// Session token entropy straight from the kernel CSPRNG. Zig 0.16
@@ -421,11 +421,10 @@ fn headerValueInt(head: []const u8, name: []const u8) ?usize {
 
 // ----------------------------------------------------------- json helpers
 
-/// Tiny extractor for the two flat shapes the hooks send. Not a JSON
-/// parser: finds "key":"value" (no escape support needed, states are
-/// enum words and bubble text is capped ASCII-ish) and "key":number.
-/// Escaped quotes in bubble text truncate at the escape, which only
-/// shortens the bubble, never corrupts memory.
+/// Small allocation-free parser: finds "key":"value" and "key":number.
+/// String slices retain JSON escapes for the caller to decode or flatten, but
+/// the scanner must skip them so multiline assistant messages are not cut at
+/// their first `\n`.
 pub fn jsonStringPub(body: []const u8, key: []const u8) ?[]const u8 {
     return jsonString(body, key);
 }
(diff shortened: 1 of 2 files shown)


Railly:
#626 landed first, so this needs a rebase. I would normally do it for you (I rebased #614's siblings and several others this week), but this one has three conflicts where the right answer is a product decision rather than a mechanical merge, and guessing wrong would silently undo work.

Ten conflicts total in `main.zig`. Seven are keep-both: new `Msg` variants, `Model` fields, load and apply paths, the settings JSON line. Those I would just take both sides of.

The three that need you:

**1. The settings buffer.** #626 raised it to 448 with the worst case measured in a comment. This PR sets 384. Taking your side shrinks a buffer that now has more fields to hold, and a `bufPrint` overflow there fails silently and drops the entire save. The merged line needs `bubble_lifetime`, `bubble_columns` and `bubble_answer_lines` together, so 448 or higher.

**2. The spinner slot.** #626 added a static `!` marker while `model.state == .waiting`, so a permission prompt is visible without waking the frame loop. This PR replaces that block with the busy-or-empty version and drops the marker. Your version also parameterizes the size via `bubble_busy_width`, which #626 does not have. Both changes are wanted; they need combining rather than picking.

**3. The constants block.** #626 added `bubble_lifetime_*` next to `bubble_text_*`; this PR replaces the region with `bubbleMaxCardWidth`. Both belong.

Everything else looked good on inspection, and the split into #625 and #626 is exactly what I asked for on #614: no SDK patch, no workflow changes, based on main. #626 is already in, verified at 39/39 with green macOS and Windows builds.

Push the rebase and I will verify and merge. If you would rather I take it, say so and I will make those three calls in your favor and tell you exactly what I chose.

ThermalEng:
Thanks — please take the rebase. Your proposed resolutions are exactly what I want:

- keep the settings buffer at 448 bytes or larger;
- keep #626’s static waiting `!`, using #625’s `bubble_busy_width` for the shared spinner/marker slot;
- keep both the `bubble_lifetime_*` constants and #625’s bubble sizing/layout constants and `bubbleMaxCardWidth`;
- keep both sides for the remaining mechanical conflicts.

Please go ahead and verify and merge with those choices. Thank you.

Railly:
Rebased and merged with the four choices you approved. Authorship unchanged, both commits intact, +372/-116.

Two things went beyond the three I flagged, so they are worth naming:

**`bubbleCharsPerLine` had to go.** Keeping both sides resurrected it, but this PR deliberately replaces it with configurable columns and nothing calls it anymore. Dead code that still compiles is the worst kind, so I dropped it.

**The text-size bounds collided.** Main had `bubble_text_min_px = 13`, yours lowers it to 8. I kept yours, since lowering the minimum is a stated goal of the PR, and deleted main's duplicate block.

Also, your font commit raises the buffer to 1536 while #626 had it at 448, and its JSON line omitted `bubble_lifetime`. Merged line carries all of it now: `waiting_sound`, `bubble_text`, `bubble_lifetime`, `bubble_columns`, `bubble_answer_lines`, `font_path`, `hide_dock`, `rotate_pets`, `rotation_day`. Worst case with a 250-char font path measures 593 against 1536.

Verified: 41/41 tests, `native build` green on macOS and `x86_64-windows -Dcpu=baseline`, and the Linux analysis object still compiles without libc. I also checked all eight settings round-trip: written, read back, applied to the model.

Merging.

PR #626, opened 2026-07-30
feat(desktop): add configurable bubble lifetime

ThermalEng:
## Summary

- Add a shared Native desktop bubble lifetime setting in seconds, defaulting to `0` (no automatic expiry).
- Use a direct numeric input (`0–60` seconds); values from 1 to 60 enable expiry.
- Persist it in the existing `desktop-native-settings.json` as `bubble_lifetime`.
- Keep spinner/busy bubbles visible without a deadline; start or refresh the timer only for a final non-busy message.
- Keep permission and input prompts visible while the existing agent state is `waiting`.
- Show a static `!` in the spinner slot while `waiting` to make approval/input requests visible without continuous redraw.
- Clear immediately when messages are disabled, Focus Mode is enabled, an explicit empty message arrives, or the configured lifetime expires.
- Reuse the existing declarative bubble-window lifecycle so the SDK closes the secondary window when the bubble becomes inactive.

The lifetime behavior is in the shared Native desktop path and applies to Linux, Windows, and macOS.

## Scope and independence

- Based directly on upstream `main` (`d0d2c38`).
- Does not include or depend on #614, #625, or Native SDK changes.
- Changes only `packages/petdex-desktop-native/src/main.zig`.

## Validation

- `native test`: 39/39 passed against the pinned, unmodified Native SDK `65d4885`.
- Linux `native build -Dcpu=baseline -Dtrace=off`: passed (ReleaseFast).
- Windows cross-build `native build -Dtarget=x86_64-windows-gnu -Dcpu=baseline -Dtrace=off`: passed (ReleaseFast).
- `git diff --check`: passed.

### DIFF
--- a/packages/petdex-desktop-native/src/main.zig
+++ b/packages/petdex-desktop-native/src/main.zig
@@ -134,6 +134,7 @@ pub const Msg = union(enum) {
     toggle_waiting_sound,
     chime_done: native_sdk.EffectExit,
     set_bubble_text_size: f32,
+    bubble_lifetime_input: canvas.TextInputEvent,
     toggle_hide_dock,
     toggle_launch_at_login,
     toggle_focus_mode,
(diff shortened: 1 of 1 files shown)


Railly:
Thank you for splitting the work. This is exactly the shape I asked for on #614: one file, no SDK patch, no workflow changes, based on main.

Verified independently rather than taking the validation section:

- `native test`: 39/39, matching your number exactly
- `native build` green on macOS
- `native build -Dtarget=x86_64-windows -Dcpu=baseline` green
- diff touches only `src/main.zig`, confirmed against main

I also checked the settings buffer, since a `bufPrint` overflow there fails silently and drops the whole save. Worst case with a 63-char slug and every field present comes to 282 bytes against your 448, so 166 to spare. Growing it alongside the new key was the right call.

Two design details worth naming, both correct: defaulting to `0` means existing installs see no behavior change, and keeping spinner and `waiting` bubbles alive without a deadline avoids the obvious failure where a permission prompt expires before anyone reads it.

Merging.

PR #628, opened 2026-07-31
fix(desktop): grep bubbles lose their pattern to an ASCII quote

sorrycc:
Every Grep bubble renders as a bare `Searching ` — the pattern never makes it to the screen. This has been true on every agent since the in-binary runner landed, and it reproduces on a stock install:

```
$ echo '{"tool_name":"Grep","tool_input":{"pattern":"petdex"}}' \
    | ~/.petdex/bin/petdex-hook bubble pre claude-code
$ curl -s localhost:7777/bubble
{"text":"Searching ", ...}          # want: Searching “petdex”
```

## Why

`hook_runner.run` interpolates the bubble text raw into the POST body:

```zig
bufPrint(&body_buf, "{{\"text\":\"{s}\",\"busy\":{},\"agent_source\":\"{s}\"}}", .{ text, busy, agent })
```

and `hook_server.jsonString` reads it back with a scanner that stops at the first `"` **or** `\` and decodes neither:

```zig
while (i < body.len and body[i] != '"' and body[i] != '\\') i += 1;
```

Every other template honors that contract — payload-derived slices come out of the same scanner, so they carry no bare quote and no stray backslash. Grep is the one template that adds ASCII quotes of its own, which closes the JSON string early and drops the rest.

Escaping to `\"` does not fix it. I checked against the running sidecar:

| body sent | text read back |
| --- | --- |
| `{"text":"Searching "TODO""}` | `Searching ` |
| `{"text":"Searching \"TODO\""}` | `Searching ` |
| `{"text":"Searching “TODO”"}` | `Searching “TODO”` ✅ |

So the fix is curly quotes — plain bytes that survive both steps, the same way `Thinking…` and `Waiting for you…` already do.

## Scope

- **`hook_runner.zig`** — the actual user-visible fix.
- **`assets/opencode-plugin.js`** — same template. It serializes with `JSON.stringify`, so its body is well-formed, but the reader still stops at the backslash and the bubble truncates identically. `scan()` compares the installed plugin against the embedded copy byte-for-byte, so existing installs will surface as *Plugin outdated* and **Update** refreshes them.
- **`bubble-templates.ts`** — the parity source the Zig port is kept against. Moved in step even though the CLI no longer dispatches `bubble`, so the two do not drift.

Deliberately not touched: `jsonString` itself. Teaching it to decode escapes is a change to the parser every hook payload and every runtime mirror file goes through — a much larger blast radius than this bug warrants.

## Tests

A Grep case on each side, plus a round-trip test that pushes every template through the real body format string and back through `hook_server`'s reader:

```zig
const body = try std.fmt.bufPrint(&body_buf, "{{\"text\":\"{s}\",...}}", .{text});
try t.expectEqualStrings(text, jsonString(body, "text").?);
```

Both new tests fail against the old templates and pass against the new ones — I reverted the templates to confirm the assertions have teeth, not just green.

```
$ zig test hook_runner.zig     # 0.16.0
All 11 tests passed.
$ bun test                     # packages/petdex-cli
74 pass, 0 fail
```

`agent_hooks.zig` (18), `installer.zig` (5) and `plat.zig` (1) also still pass. Verified end to end against a live `Petdex.app` 0.3.1 sidecar on macOS 26.3 (arm64).

### DIFF
--- a/packages/petdex-cli/src/hooks/bubble-templates.test.ts
+++ b/packages/petdex-cli/src/hooks/bubble-templates.test.ts
@@ -144,7 +144,29 @@ describe("formatBubble - grep + glob", () => {
         toolName: "Grep",
         toolInput: { pattern: "TODO" },
       }),
-    ).toBe('Searching "TODO"');
+    ).toBe("Searching “TODO”");
+  });
+
+  test("no template renders an ASCII quote", () => {
+    // The in-app runner embeds this text raw into the sidecar POST body
+    // and hook_server reads it back with a scanner that decodes nothing,
+    // so a `"` or a `\` anywhere in a template truncates the bubble.
+    for (const toolName of ["Grep", "Glob", "Read", "Bash", "WebFetch"]) {
+      for (const phase of ["running", "done"] as const) {
+        const text = formatBubble({
+          kind: "tool",
+          phase,
+          toolName,
+          toolInput: {
+            pattern: "TODO",
+            file_path: "a.ts",
+            url: "https://x.dev",
+          },
+        });
+        expect(text).not.toContain('"');
+        expect(text).not.toContain("\\");
+      }
+    }
   });
 
   test("Glob done with pattern", () => {
--- a/packages/petdex-cli/src/hooks/bubble-templates.ts
+++ b/packages/petdex-cli/src/hooks/bubble-templates.ts
@@ -110,9 +110,11 @@ export function formatBubble(event: BubbleEvent): string {
     case "grep": {
       const pattern = fieldFrom(toolInput, "pattern");
       if (pattern) {
+        // Curly quotes keep this in parity with the in-app runner, where
+        // an ASCII `"` truncates the bubble on the way to the sidecar.
         return past
-          ? `Searched "${clip(pattern, 28)}"`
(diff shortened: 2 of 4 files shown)


Railly:
Confirmed the bug before trusting the diagnosis. Simulated the scanner against the body main actually produces:

```
body sent: {"text":"Searching "petdex"","busy":true,...}
text read: 'Searching '
expected:  'Searching "petdex"'
```

Exactly what you describe. `hook_runner.zig:183` wraps the pattern in ASCII quotes and line 100 interpolates it raw, so the value terminates early and the pattern disappears. Grep is the only template that adds quotes of its own, which is why nothing else broke.

Typographic quotes are the right fix rather than escaping. The scanner skips escape pairs but does not decode them, so `\"` would render literally. Sidestepping the character entirely keeps the wire format honest and reads better in the bubble.

Verified after rebase: 43/43 native tests, 60/60 CLI hook tests, `native build` green on macOS and `x86_64-windows -Dcpu=baseline`, and the scanner reads `Searching “petdex”` back intact.

Nice first contribution, and a good catch: this shipped broken on every agent since the in-binary runner landed and nobody noticed. Merging.

PR #629, opened 2026-08-01
feat(desktop): add Qoder as an agent, and light the failed sprite row

sorrycc:
image

Google Chrome 2026-08-01 10 23 26

And this is the detailed design doc for refence, https://gist.github.com/sorrycc/837b51198f135d7a0438feb25e4712ff

---

Adds **Qoder** to the Agents panel. Its hook system is Claude Code's shape — same `settings.json` layout, same tool names, `Agent` where Claude says `Task` (which `formatBubble` already matched). **Existing bubble templates needed no changes.**

The interesting part is the event Claude Code doesn't have.

## It lights the `failed` row

Petdex has shipped a `failed` sprite row since v1 — row 5, 8 frames, 1220ms — that **no agent has ever lit**, because no connected agent reports tool failure. Qoder emits `PostToolUseFailure`, so this adds a `tool-failure` phase that maps to `failed`.

The two Post\* events don't fight: `coreToolHookTriggers.ts:288` gates `PostToolUse` on `!toolResult.error`, so a failing call emits `PostToolUseFailure` *instead of* `PostToolUse` and no trailing `idle` stomps the state.

Event map: `UserPromptSubmit`→jumping · `PreToolUse`→review/running · `PostToolUse`→idle · **`PostToolUseFailure`→failed** · `Notification`→waiting · `Stop`→waving. `SubagentStart/Stop` and `PermissionRequest` are deliberately not wired — concurrent subagents fight over one global pet state, and `PermissionRequest` duplicates `Notification`.

## One row, several config roots

Qoder ships a global build (`~/.qoder`) and a CN build (`~/.qoder-cn`) that coexist on one machine, and the same `settings.json` drives both its IDE and its CLI — hence `Qoder`, not `Qoder CLI`.

I tried two `AgentKind`s first (the `CLAUDE_CONFIG_DIR` / #601 argument: separate installs deserve separate control). On screen it was just noise for one product. So both roots sit behind one row:

- Install covers every root present; Disconnect clears every root present
- Status folds **worst-first** — any root still needing a press decides the button, so one Install always completes every root
- `qoder_roots` is a table rather than a `cn` boolean, so resolve/scan/install/uninstall are one loop and a third brand would be one entry

**A root whose `settings.json` could never be merged into is excluded, not counted as unhooked.** `classifyConfig` reports a malformed config as `.none` while `installJsonHooks` refuses to touch it — folded together that would peg the row at "Hooks not installed" behind a button that can never succeed. The refused config stays byte-for-byte untouched.

Env overrides are per brand: `*_CONFIG_DIR` (a whole root, like `CLAUDE_CONFIG_DIR`) and `*_CLI_HOME` (replaces the home dir instead). Skipping `*_CLI_HOME` would write hooks where that user's Qoder never reads. `*_CONFIG_DIR_NAME` is out of scope — it renames only the leaf and degrades to a visible "Not detected" rather than a wrong-root write.

Install writes **nothing outside `hooks`**: Qoder's `hooksConfig.enabled` already defaults `true` upstream, so writing it (as we do for Gemini) could only clobber a deliberate opt-out.

## Two changes worth a second look

**`stateBody` extracted as a pure function on both sides.** The `/state` body was built inline in `run()` / `runBubble()`, behind a stdin drain, a token read and a socket — so nothing could test it, which is why all the existing tests are pure. It's now one format string with an optional fragment, so `duration_ms == 0` *provably* renders exactly the bytes that shipped before this phase existed, rather than that being a promise someone has to keep across future edits.

**The failure bubble is the tool name only** — `Bash failed`, never the payload's `error`. `hook_server.jsonString` stops at the first `"` or `\` and decodes neither, and error strings routinely carry both. `tool_name` is a controlled identifier from Qoder's own registry, so this satisfies the constraint structurally rather than by careful wording.

## Deliberately not here

- **The general `dwellFor` fix.** `failed` is 1220ms of animation but `dwellFor(.failed, 0)` returns the 250ms floor, so the `tool-failure` `/state` POST carries an explicit `duration`. Making `dwellFor` fall back to each duration state's intrinsic length is the honest fix and repairs `waving`/`jumping`/`review` too — but it would stretch Claude Code's and Codex's Stop wave from 250ms to 1120ms, a visible change for every shipped user in a PR about adding an agent. Happy to send it separately.
- **`is_interrupt` handling.** No caller passes it to `firePostToolUseFailureEvent` (`hookEventHandler.ts:471`, `:485`), so the field is never emitted; reading it would mean adding a bool reader for a key that never arrives.
- **README changes.** Neither README enumerates agents — a case-insensitive grep for claude/gemini/opencode returns zero hits in both — so there's no list to append to.
- **New i18n keys.** The agent-enumerating prose is updated in all three locales (values only); SEO keyword entries felt like a maintainer's call.

## Artwork

`assets/agents/qoder.png` is the bare glyph on transparent, 40×40 RGBA. I started from the official dark-tile app icon and backed out: against the dark `.surface` (`rgb8(25,25,28)`, `main.zig:287`) a `#111113` tile computes to **1.08:1**, and against the light one **18.86:1** — invisible in one theme, dominant in the other, in a list whose rows should read as peers. Every existing tile in the set is light or chromatic (`codex.svg` is `#fff`, `gemini.svg` a gradient); none is near-black. The bare glyph has a fully transparent lens, so one asset works in both themes and `agent_art`'s identical `{light, dark}` pair stays truthful. `public/brand/agents/qoder.svg` keeps the official vector for the homepage, where it renders at 16px with `opacity-80` alongside codex's and gemini's tiles.

## Verification

- `zig test agent_hooks.zig` **18 → 27**, `zig test hook_runner.zig` **9 → 12** (Zig 0.16.0, the CI pin). New coverage: the worst-first fold as a 3×3 table asserted in both directions, install/disconnect across every root, the excluded-broken-root case, same-path dedup, per-root env isolation, and a round-trip that pushes the new template through the real body format string and back through `hook_server`'s reader. Fixture tests are re-run-safe.
- `packages/petdex-cli`: `bun test` **74 → 78**, `bun run build && bun run typecheck` clean, `bun.lock` untouched.
- Root: `bun run check` and `bun run i18n:check` pass.
- Manually: built via the pinned SDK, connected from Settings, and confirmed one Install writes six events into **both** roots (12 hooks) while preserving the foreign hooks already in each and leaving `hooksConfig` alone.

`bun run build` at the root fails on `TypeScript 7.0.2 does not provide the compiler API required by Next.js` — I reproduced that on a clean `main`, so it predates this branch.


Railly:
Rebased onto main for you (#625 and #628 landed in between). Authorship unchanged, +730/-28.

One conflict, in `agent_icon_ids`. I took your side and checked the claim rather than the comment: slots in use are 1, 9, 10, 11, 13, 14, 15, so 16 is genuinely the last one free against the 16-slot cap. Worth flagging that the next agent after Qoder has nowhere to go, which is a good problem to hit and a bad one to hit by surprise.

Verified: 53/53 native tests, 336/336 web tests, `i18n:check` in sync across en/es/zh, `native build` green on macOS and `x86_64-windows -Dcpu=baseline`.

Lighting the `failed` row is the part I like. A sprite row that shipped in v1 and no agent ever triggered is dead weight that looks like a feature, and your reasoning for why the two Post\* events cannot both fire (the `!toolResult.error` gate) is the detail that makes it safe rather than racy.

Leaving `SubagentStart/Stop` and `PermissionRequest` unwired is the right call too. Concurrent subagents fighting over one global pet state would show up as flicker nobody could explain.

Merging.

PR #637, opened 2026-08-02
fix(desktop): resize the bubble window with its content, outline it in light mode

Railly:
Two bugs found by looking at the running app, both reported with screenshots.

## The bubble grew but its window did not

Raising **Bubble text size** made the card wider while the window kept the size it was created at. The card then rendered past the window's right edge and the tail pointed at empty space instead of the pet, so the whole bubble looked shoved to the left.

Root cause: `resizeWindow("bubble", ...)` appeared nowhere in the file. The window is created once in the shell descriptor and never resized. The three settings that change bubble geometry (`set_bubble_text_size`, `bubble_columns_input`, `bubble_answer_lines_input`) all called `fitWindow`, which resizes **the pet's** window, not the bubble's.

`syncBubbleWindow` now resizes before it moves. The order is load-bearing: the move centers on the new width, so resizing after it would leave the bubble offset by half the delta, which looks like a smaller version of the same bug.

## The tail vanished on light backgrounds

`registerTail` drew its diagonal hairline behind a dark-mode check, and the light card had no border set at all. A white bubble over a white editor has no silhouette, and the tail is the first thing to disappear because it is the narrowest part.

Both now carry an outline, with the tail hairline matched to the card border so the join still reads as one shape rather than two stacked pieces. Light mode needs this more than dark does, not less.

## Verified

Screenshot at 20pt, the size that reproduced it: the bubble sits centered over the pet with the tail on its midline, and the outline is visible against a light background.

65/65 tests, `native build` green on macOS and `x86_64-windows -Dcpu=baseline`.

Worth noting these were only visible by running the app with a real pet and a real bubble. The adapter PRs earlier this week were verified by screenshot too, but of the Agents panel, which never exercises this path.

### DIFF
--- a/packages/petdex-desktop-native/src/main.zig
+++ b/packages/petdex-desktop-native/src/main.zig
@@ -1197,14 +1197,18 @@ fn registerTail(dark: bool, fx: *Effects) void {
                 pixels[i + 1] = cg;
                 pixels[i + 2] = cb;
                 pixels[i + 3] = 255;
-                // The card carries a hairline in dark mode: the tail's
-                // diagonal edges continue it so the join reads as one
-                // outlined shape (the top row stays plain - it tucks
-                // under the card).
-                if (dark and y > 0 and (x == inset or x == tail_w - inset - 1)) {
-                    pixels[i] = 48;
-                    pixels[i + 1] = 48;
-                    pixels[i + 2] = 53;
+                // Hairline along the tail's diagonal edges so the join
+                // reads as one outlined shape (the top row stays plain,
+                // it tucks under the card). Both themes need it: a white
+                // tail on a light desktop background has no silhouette
+                // at all, which is the same reason the card is outlined.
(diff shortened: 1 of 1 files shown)


Railly:
Third fix pushed to the same branch: the bubble text default.

The default read `bubble_text_min_px`, which was 13 when the slider shipped in #610. #625 lowered the range floor to 8, a reasonable change for anyone who wants a denser bubble, but left the default pinned to the minimum. So every install silently shrank to 8pt and the slider sat hard left.

The default is its own constant now, which means widening the range again cannot move what a fresh install looks like. A test pins it away from both bounds so the next range change does not repeat this.

Verified on a scratch HOME: a fresh install persists `bubble_text: 13.0`, the slider lands at 42 percent, and the bubble reads at the size it did before #625.

All three fixes in this PR came from running the app rather than from tests, and none of them would have failed CI.

PR #654, opened 2026-08-03
修复 CLI petdex edit 的 Bearer 编辑接口

MisonL:
## 概要

完善 `petdex edit` 的 Bearer 编辑链路和待审核资产生命周期，确保 CLI 编辑与 Web 编辑使用一致的审核边界，并避免无效上传、孤儿对象和并发清理造成状态损坏。

## 变更

- 保留并完善 Bearer 专用 `PATCH /api/cli/edit` 路由：复用 Pet 所有者校验、内容审核、限流、自动审批和缓存失效逻辑。
- `petdex edit` 在上传前校验本地资产：严格解析 PNG/WebP（含动画 WebP）、JSON 和 ZIP，校验真实文件格式、尺寸、大小和 MIME 类型；预签名或上传失败会显式终止，不再静默跳过。
- 统一提交和编辑资产的 8 MB 限制与多语言提示，预签名请求和实际提交分别限流；编辑预签名额度按用户维度计算，避免通过多个 Pet 绕过孤儿对象预算。
- 生成严格的提交/编辑资产 key，并在服务端仅接受当前 Pet 对应的 pending 资产 URL，阻止跨 Pet、错误扩展名、查询串和非受信主机引用。
- 修复重复编辑的 pending 状态合并、无操作判断和审核指纹清理；`spriteVersionNumber` 已贯通 `PatchBody`、pending patch、无操作判断、资产更新 SQL 和 CLI metadata 编辑路径。
- pending 资产写入使用显式事务：先单独获取 advisory lock，再读取 claim 并更新 Pet。Neon HTTP 使用事务 batch，本地 Postgres/PGlite 使用 callback transaction，避免锁等待前的快照竞态。
- 新增 pending R2 GC：默认 dry-run，`--apply` 才执行删除；通过数据库 claim 表和 advisory lock 与编辑写入串行化。R2 未返回某个 `Deleted` 项且没有显式错误时按幂等成功处理，显式 R2 错误仍失败。
- GC 在成功或幂等成功后释放 claim，并在 apply 模式清理已不存在且未引用的过期 claim；失败 claim 保留以便后续重试。
- 将 pending claim 表迁移重命名为 `0020_pending_asset_gc_claims.sql`，避开上游已经使用的 `0018_sprite_version_number.sql`。
- 增加覆盖边界条件的单元测试。

## 根因

原 CLI 编辑流程把未经验证的本地文件直接上传，并在预签名接口异常时可能继续提交；服务端 pending 资产校验也无法区分当前 Pet 的合法待审核对象。与此同时，pending R2 清理与编辑写入之间存在快照竞态，静默删除响应还会把幂等缺失误报为失败，导致状态与对象存储出现不一致风险。

## 验证

- `bun run check`（退出码 0；仅有既有可修复提示）
- `bun test --timeout 60000`：394 个测试通过，0 个失败
- `bunx tsc --noEmit`
- `packages/petdex-cli`：`bun run build`、`bun run typecheck`
- `PETDEX_MOCK=1` 搜索和 Pet 映射回归测试
- `git diff --check`

本 PR 已 rebase 到上游最新 release 基线，不包含本机路径、内网地址、Docker 配置、访问令牌或其他环境信息。


Railly:
Reviewed this properly. The Bearer edit endpoint itself is sound, and the auth boundary is right: `userId` comes only from `verifyCliBearer`, never from the body, and your own test proves it. `isPendingAssetUrl` and `buildPendingAssetKey` are solid too, anchored regexes and a rejected host list, no bypass found.

Four things need to change before this can merge.

**1. The claim guard does not prevent the race it is built for.** `src/lib/pet-edit.ts:425-456` acquires `pg_advisory_xact_lock` and reads `pending_asset_gc_claims` in a single statement. In READ COMMITTED the statement snapshot is taken before the lock returns, so the `NOT EXISTS` reads a pre-commit view. Reproduced on Postgres 14: the GC claims a key, the concurrent edit still returns `UPDATE 1` instead of 0, and the queued edit ends up pointing at an object the GC then deletes. The lock works, the visibility does not. Splitting the lock into its own statement inside an explicit transaction makes the follow-up read see the claim.

**2. This branch drops `spriteVersionNumber`, which shipped in #658.** That feature lives inside `src/app/api/my-pets/[id]/edit/route.ts`, the file this PR reduces to 6 lines, and the extracted logic does not carry it: it is missing from `PatchBody`, from `PendingEditPatch`, from the no-op check, and from the hand-written column list in the raw `UPDATE`. That last one also leaves `pending_sprite_version_number` stale while the rest of the pending row is rewritten. The conflict here is semantic, so resolving it by taking this side would silently revert #658.

**3. `summarizeR2DeleteBatch` (`src/lib/r2.ts`) treats a missing key as a failure.** Deleting a key that no longer exists is an idempotent success, and R2 does not guarantee a `Deleted` entry for it. As written, `gc-pending-r2.ts:194` sets `process.exitCode = 1`, so the GC job would fail red routinely, and the claim is already persisted at that point, so the key never gets retried.

**4. Renumber the migration to `0020`.** `0018` is now taken by `0018_sprite_version_number.sql` on main. The table itself is fine: new table plus index, additive, does not touch `submitted_pets`.

Two smaller ones, worth doing before any cron runs this: `pending_asset_gc_claims` has no purge path (the `claimed_at` index is never read), and the presign rate limit is keyed `userId:petId`, so the budget is per pet rather than per user while the orphans it bounds are global.

A suggestion on shape. This PR is four independent changes: the Bearer endpoint fix, the R2 GC, rate limiting, and the presign rework. The endpoint fix is the part the title promises and the part that is healthy. If you split it out, it can merge now while the GC and the claim race get their own PR with room to fix the visibility problem properly.

It also needs a rebase; main has moved a fair bit today.

Thanks for the work here. The GC design is careful, dry-run by default with an explicit `--apply` and a serializable claim, and the race is a subtle one that only shows up under real concurrency.

MisonL:
Thanks for the careful review. I addressed all requested items in this PR:

- Split the advisory-lock acquisition from the asset update inside the same transaction. The Neon batch path and the local callback transaction both execute the UPDATE after the lock, so it reads a fresh snapshot.
- Preserved `spriteVersionNumber` through `PatchBody`, pending patch construction, no-op detection, the raw UPDATE, and the CLI metadata edit path, including `pending_sprite_version_number`.
- Treat an omitted R2 `Deleted` entry without an explicit `Error` as idempotent success; explicit per-key errors still fail.
- Renumbered the claim migration to `0020_pending_asset_gc_claims.sql`.
- Added apply-only cleanup for expired, unreferenced claims. Claims are released after confirmed or idempotent deletion, while failed deletions keep their claims for retry.
- Changed the edit presign limit to be keyed by `userId`, so the budget applies across all pets owned by one user.

Verification completed:

- `bun run check`: passed (one existing informational Biome suggestion only)
- `bun test --timeout 60000`: 394 passed, 0 failed
- `bunx tsc --noEmit`: passed
- CLI package `bun install --frozen-lockfile`, build, typecheck, and tests: 92 passed, 0 failed
- Packaged CLI version check and local `npx` smoke test: passed
- `git diff --check`: clean

The repository CI checks pass across macOS, Linux, and Windows with Node 20, 22, and 24. The only remaining non-code check is Vercel team authorization.

MisonL:
Hi @Railly, a production smoke check after the latest 1.2.0 release still shows:

- `POST /api/cli/edit-presign` returns `401` JSON, so the presign route is deployed.
- `PATCH /api/cli/edit` returns `404` HTML, so the Bearer edit submission route from #654 does not appear to be deployed yet.

Could you please merge/rebase and deploy #654, or confirm the intended production submission route? No duplicate edit submission was attempted.

Railly:
Merged. Rebased onto main and re-verified: 396 app tests, 94 CLI tests, tsc and biome clean, 9/9 CI jobs green.

Checked all six review items against the code. The lock split is correct, each statement takes its own snapshot so the `NOT EXISTS` sees the committed claim. `spriteVersionNumber` reaches the raw UPDATE, migration is `0020`, missing R2 keys count as deleted, presign limit is keyed by user.

Your 404 diagnosis was right: `/api/cli/edit` did not exist on main. It ships with this merge.

Issue #603, opened 2026-07-27
[macOS] Keep the pet and activity bubble together across Spaces and follow the active agent window


chenqing0106:
## Summary

On macOS, the native desktop app renders the pet and its activity bubble as two independent floating windows. They can become separated when switching windows, Spaces, full-screen apps, or displays. The pet also stays at a fixed absolute screen position instead of following the active coding-agent window.

This makes the activity bubble appear on one desktop/window layer while the pet remains on another, so the status and the mascot no longer feel like one UI.

## Environment

- Petdex Desktop: `0.3.0` native app
- macOS: 26.1
- Architecture: Apple Silicon (`arm64`)
- Agent: Codex Desktop / Codex CLI hooks

## Steps to reproduce

1. Launch Petdex Desktop and place the pet next to a Codex or terminal window.
2. Trigger a tool call so the activity bubble appears.
3. Switch to another app window, macOS Space, full-screen window, or another display.
4. Continue triggering agent activity.

## Actual behavior

- The pet and bubble may appear on different Spaces or window layers.
- The bubble can remain visible over the desktop while the pet is over a different application window.
- Switching to another Codex/Terminal/IDE window does not move the pet with the active workflow.
- The pet remains at its previous absolute screen coordinates.

## Expected behavior

At minimum, the pet and bubble should behave as one window group:

- stay together across Space and full-screen transitions;
- share the same always-on-top behavior;
- move together when the pet is dragged;
- remain on the display/Space containing the pet.

It would also be useful to offer an optional **Follow active agent window** setting. When enabled, switching between supported agent windows (Codex, Terminal, iTerm, Warp, VS Code, etc.) would place the pet beside the focused window and move the bubble with it.

Possible modes:

1. **Fixed on screen** — current absolute-position behavior.
2. **Follow current Space** — visible on every Space without changing its relative screen position.
3. **Follow active agent** — move beside the focused supported agent window.

## Source observation / possible implementation

`packages/petdex-desktop-native/src/main.zig` explicitly creates the bubble in its own fixed-size window. Both the main pet window and bubble window use `.floating = true`, but they are not represented as a parent/child window group and no all-Spaces/full-screen collection behavior is declared.

On macOS, possible implementation pieces are:

- keep both windows at the same `NSWindow.Level`;
- use `addChildWindow(_:ordered:)` (or an equivalent native-SDK parent-window relationship) so the bubble follows the pet;
[...]

PR #656, opened 2026-08-04
feat(desktop): focus originating terminal and support fullscreen overlays

### LINKED ISSUE #596: feature: focus the originating terminal when clicking the pet
## Summary

Allow clicking the Petdex Desktop pet to focus the terminal or coding-agent window that triggered its latest state update.

## Motivation

Petdex can react to activity from multiple coding agents and terminal sessions, but the pet currently acts only as a visual status indicator.

When a task finishes, fails, or waits for input, the user still has to manually find the relevant terminal window or tab. This becomes especially difficult when running Codex, Claude Code, Gemini CLI, OpenCode, or other agents in parallel.

Clicking the pet could provide a direct way back to the session that needs attention.

## Proposed behavior

For an initial implementation:

1. Hook state updates include origin metadata where available:
   - `agent_source`
   - `session_id`
   - `cwd`
   - terminal application
   - PID / TTY / window or tab identifier
2. The sidecar remembers the origin of the latest relevant state update.
3. Clicking the pet focuses the originating terminal or coding-agent application.
4. If an exact window or tab cannot be identified, Petdex falls back to activating the originating application.
5. If no valid origin is available, clicking the pet keeps its current behavior.

For multiple active sessions, the most recently updated session could own the click target as an MVP. Per-session or multi-pet routing could be explored separately.

## Example

1. A Codex session in iTerm2 enters the `waiting` state.
2. Petdex displays the waiting animation.
3. The user clicks the pet.
4. Petdex brings the corresponding iTerm2 window or tab to the foreground.

## Acceptance criteria

- Clicking the pet can focus the application associated with the latest state update.
- Existing dragging and right-click behavior continue to work.
- Missing or stale process/window information is handled gracefully.
- Hook events without origin metadata remain backward compatible.
- Origin metadata stays local and is not sent to the Petdex service.
- The initial platform scope can be macOS, with other platforms added later.

## Implementation considerations

Terminal applications expose different automation capabilities. Exact tab-level focusing may require application-specific adapters for Terminal.app, iTerm2, Warp, VS Code, and Cursor.

A staged implementation could start with application-level activation, then add exact window/tab focusing for supported terminals.

Would this behavior fit the direction of Petdex Desktop? I would be interested in working on an implementation after validating the preferred scope.

chenqing0106:
## Summary

Fixes #596 and #603.

- keep the pet and bubble together across macOS Spaces and fullscreen apps
- add an explicit fullscreen overlay opt-in for the Petdex windows
- record the originating terminal metadata for supported hook events
- clicking the pet focuses the originating Apple Terminal tab by TTY, or activates VS Code
- keep the activation-policy decision consistent with `hide_dock` and fullscreen overlay mode

## Dependency

This uses the new Native SDK window option from [Railly/native#2](https://github.com/Railly/native/pull/2).

## Validation

- focused Native app test: 68/68 passed
- built and packaged the macOS app locally
- manually verified fullscreen visibility and VS Code activation on a second display

The source application list is intentionally allowlisted to Apple Terminal and VS Code for this first version.

### DIFF
--- a/packages/petdex-desktop-native/app.zon
+++ b/packages/petdex-desktop-native/app.zon
@@ -34,6 +34,7 @@
                 .resizable = false,
                 .titlebar = "chromeless",
                 .floating = true,
+                .fullscreen_overlay = true,
                 .transparent = true,
                 .restore_state = false,
                 .restore_policy = "center_on_primary",
--- a/packages/petdex-desktop-native/src/assets/opencode-plugin.js
+++ b/packages/petdex-desktop-native/src/assets/opencode-plugin.js
@@ -36,6 +36,15 @@ function clip(text, max = 40) {
   return text.length <= max ? text : text.slice(0, max - 1) + "…";
 }
 
+function originMetadata() {
+  const sourceApp = process.env.TERM_PROGRAM;
+  if (sourceApp !== "Apple_Terminal" && sourceApp !== "vscode") return {};
+  return {
+    source_app: sourceApp,
+    source_cwd: process.cwd(),
+  };
+}
+
 function canonicalToolKind(toolName) {
   const lower = String(toolName || "").toLowerCase();
   if (lower === "read") return "read";
@@ -137,7 +146,9 @@ async function notify({ state, duration, text, title, busy }) {
     duration != null
       ? { state, duration, agent_source: "opencode" }
       : { state, agent_source: "opencode" };
-  const bubbleBody = { text, agent_source: "opencode" };
+  const metadata = originMetadata();
+  Object.assign(stateBody, metadata);
+  const bubbleBody = { text, agent_source: "opencode", ...metadata };
   if (title) bubbleBody.title = title;
   if (busy !== undefined) bubbleBody.busy = busy;
   await Promise.all([
--- a/packages/petdex-desktop-native/src/hook_runner.zig
+++ b/packages/petdex-desktop-native/src/hook_runner.zig
@@ -29,7 +29,7 @@ const post_poll_ms: u64 = 5;
 
 /// argv tail after "bubble": [phase, agent?]. Reads stdin, formats,
 /// POSTs bubble + state to the sidecar. Never fails outward.
-pub fn run(phase: []const u8, arg_agent: ?[]const u8, home: []const u8) void {
+pub fn run(phase: []const u8, arg_agent: ?[]const u8, origin_app: plat.OriginApplication, source_cwd_raw: ?[]const u8, home: []const u8) void {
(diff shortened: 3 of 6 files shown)


Railly:
amazing, will check it out later today! thanks! @chenqing0106

MisonL:
Thanks for implementing #603/#596! These changes are very helpful for fullscreen focus and multi-conversation bubbles. Before merging, could you please double-check the items below? The goal is to avoid unintended effects on regular or settings windows in fullscreen or multi-display setups, and to prevent focus or bubble mismatches between concurrent sessions.

- `fullscreen_overlay` should remain off by default at the SDK level and be enabled explicitly only for the pet and bubble windows. If it is applied globally, regular `floating` windows or the settings window may inherit unexpected Space and fullscreen behavior.
- Please confirm the merge rules with the `hide_dock` activation strategy from #611 and respect the user choice `hide_dock=false`. Otherwise, the overlay may unconditionally switch the app to accessory mode or hide the Dock, especially when both switches are enabled.
- #657 maintains multiple bubbles by conversation. When clicking the pet to focus the originating app, please select the valid conversation by the bubble's latest `counter` instead of assuming the last array item is the newest. It would also be helpful to cover evicted sessions, missing `session_id`, and missing source metadata, so clicks do not focus an old session or the wrong window.

If possible, please add regression coverage for a normal Space, cross-application fullscreen, monitor switching, and the multi-conversation combinations above. This should help confirm that the fix preserves the current behavior without introducing window-display or focus regressions. Thank you!

chenqing0106:
Addressed the pre-merge review points in commit `e071432`:

- Enabled `fullscreen_overlay` explicitly on the pet shell window; settings remains a regular window and bubble overlay remains explicit.
- Removed the overlay-forced Dock policy path: `hide_dock=false` now keeps the regular activation policy, while `hide_dock=true` remains user-controlled.
- Changed originating-app selection to use the bubble with the greatest `counter`, rather than assuming the last array slot is newest.
- Added a regression test for counter order differing from array order.

Native test suite: `104/104` passed. Manual validation is still needed for normal Spaces, cross-application fullscreen, monitor switching, and concurrent-session focus.

Railly:
Verified on macOS 26.3.1 with the merged build.

The pet window's Core Graphics layer goes from 3 (`NSFloatingWindowLevel`) on main without this change to 1000 (`NSScreenSaverWindowLevel`) with it, which matches what you described in #603.

Confirmed visually too: with TextEdit in full screen (1512x949 at 0,33, no menu bar and no Dock), the pet renders on top of it.

Thanks for prototyping before asking, and for keeping it default-off and scoped to the pet and bubble.

PR #675, opened 2026-08-09
fix(desktop): anchor speech bubble caret to pet

smileBeda:
- Anchor the speech-bubble caret to the pet near screen edges
 - Preserve the caret for stacked conversation bubbles
 - Flip the caret when the bubble moves below the pet
 - Fix the macOS dev restart script under Bash 3.2

 Testing:
 - 123/123 native tests pass
 - Manually verified placement around all four screen corners

### DIFF
--- a/packages/petdex-desktop-native/src/main.zig
+++ b/packages/petdex-desktop-native/src/main.zig
@@ -1173,14 +1173,16 @@ fn loadAgentsAtlas(dark: bool, fx: *Effects) void {
 }
 const tail_w: usize = 18;
 const tail_h: usize = 9;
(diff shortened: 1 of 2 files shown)


Railly:
Verified locally against the pinned SDK. 123/123 tests pass, and the three new ones actually guard: reverting `bubbleTailCenterX` to the card midpoint breaks the build, and flipping the source rect fails with `expected 9, found 0`.

Also reproduced the Bash 3.2 fix on the system shell. The old loop raises `unbound variable` with an empty array under `set -u`, the guarded version does not.

Nothing to change here.

PR #676, opened 2026-08-09
fix(desktop): wrap settings descriptions

smileBeda:
Wraps settings descriptions so long explanatory text remains readable at the default macOS window width.
Adds a regression test for wrapping behavior.

Testing:
- 120/120 tests pass
- ReleaseFast build succeeds
- Visually verified on macOS

### DIFF
--- a/packages/petdex-desktop-native/src/main.zig
+++ b/packages/petdex-desktop-native/src/main.zig
@@ -3856,6 +3856,7 @@ test {
     _ = hook_server;
     _ = installer;
     _ = plat;
+    _ = settings_view;
 }
 
 test "bubble geometry follows columns lines and font size" {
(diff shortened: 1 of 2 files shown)


Railly:
The wrapping change is right, but the test was not running. Pushed a one-line fix to your branch.

`zig build test` only collects tests from the root module file, and the aggregator block in `main.zig` lists `agent_hooks`, `hook_runner`, `hook_server`, `installer` and `plat`, not `settings_view`. Your test compiled green and never executed. I caught it by reverting `mutedParagraph` back to `ui.text`: still 120/120 pass. Setting the span assertion to an impossible 999 also passed.

With `_ = settings_view;` added it runs: 121 tests, and the revert now fails with `expected 1, found 0`.

Worth knowing for future tests in this package, an inline test in any file other than `main.zig` needs its module referenced in that block.

PR #677, opened 2026-08-09
test(cli): guard TRUSTED_ASSET_HOSTS against drift from the server allowlist

### LINKED ISSUE #664: CLI: test that TRUSTED_ASSET_HOSTS stays in sync with the server-side allowlist
## Context
`packages/petdex-cli/src/asset-hosts.ts` documents that it mirrors `src/lib/url-allowlist.ts` server-side, and warns that drift either rejects legit installs or accepts attacker bytes. That security invariant has no automated check.

## Task
Add a test (or CI script) comparing the two host sets so drift fails the build.

## Acceptance criteria
CI fails when the CLI and server allowlists diverge.

jmdlrg:
## Summary

Adds the drift guard requested in #664: the CLI's `TRUSTED_ASSET_HOSTS` is now compared against the server-side allowlist (`listAllowedHosts()` from `src/lib/url-allowlist.ts`) in `asset-hosts.test.ts`, so any divergence fails `bun test` — which both `cli-ci` and `cli-release` workflows already run.

- Exported `TRUSTED_ASSET_HOSTS` as a `ReadonlySet<string>` to make the comparison possible without widening the mutable surface.
- Two tests: strict set equality, plus a behavioral check that every server-allowed host is accepted by `isTrustedAssetUrl`.
- Note: CI runs without an `R2_PUBLIC_BASE` override, so the server list resolves deterministically to the canonical baseline the hardcoded CLI set must mirror.

Verified the guard actually guards: temporarily adding a host to the CLI set makes the test fail as expected.

Full suite passes (367 tests), `biome check` clean.

Closes #664

### DIFF
--- a/packages/petdex-cli/src/asset-hosts.test.ts
+++ b/packages/petdex-cli/src/asset-hosts.test.ts
@@ -1,15 +1,20 @@
 import { describe, expect, test } from "bun:test";
 
-import { listAllowedHosts } from "../../../src/lib/url-allowlist";
+// Imports the leaf module the server allowlist is built from, not
+// url-allowlist itself: that file pulls in the `@/` alias, which the CLI's
+// own tsconfig cannot resolve, and importing it would drag the app's module
+// graph into the published package's typecheck. `listAllowedHosts()` returns
+// `[...R2_TRUSTED_HOSTS]` verbatim, so this compares the same set.
+import { R2_TRUSTED_HOSTS } from "../../../src/lib/r2-public-url";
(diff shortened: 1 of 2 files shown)


Railly (approved):

Railly:
Pushed a fix to your branch, CI is green.

The test imported `url-allowlist`, which uses the `@/` alias the CLI tsconfig cannot resolve, so `tsc` failed while `bun test` passed. Now it imports `R2_TRUSTED_HOSTS` from `r2-public-url`, a leaf module, same set.

Verified drift fails in both directions. Thanks for adding the guard.

PR #705, opened 2026-08-15
fix(desktop): harden remote SSH runtime

skb50bd:
## Scope

- discover and invoke an absolute Windows OpenSSH client without a POSIX supervisor
- enforce private remote Petdex runtime-directory permissions
- replace managed executable symlinks instead of following an unrelated target
- retain config-symlink relocation support and extend remote tunnel regression coverage

## Why

Relative Windows SSH lookup could fall back to current-directory search behavior, permissive pre-existing runtime directories were not corrected, and following a managed executable symlink could overwrite an unrelated target.

## Impact

Windows desktops can manage POSIX SSH targets through the installed OpenSSH client. Existing POSIX desktop and remote behavior remains compatible. Managed remote state is private and symlink replacement cannot escape its intended path.

## Size

3 files changed, 182 insertions, 22 deletions.

## Validation

- `git diff --check`
- `zig fmt --check src/remote_ssh.zig`
- pinned Native SDK `native test .`: 213/213 passed
- passing GitHub Actions app and lockfile checks
- passing GitHub Actions Native SDK builds and tests on macOS, Linux, and Windows
- passing Linux real-SSH remote-agent writeback scenario

## Merge readiness

- One focused commit directly on `crafter-station/petdex:main` at `0243af6`.
- The local macOS environment limitation does not leave the tunnel unverified: the passing Linux CI job exercises the real SSH writeback path end to end.
- This remains a draft only because the Vercel preview needs Crafter Station team authorization. That is an external administrative gate; all repository checks passed.

### DIFF
--- a/packages/petdex-desktop-native/README.md
+++ b/packages/petdex-desktop-native/README.md
@@ -93,7 +93,9 @@ mirrors the desktop hook runner's contract: stdin drain, killswitch
 never fails outward. The opencode plugin POSTs directly and works unchanged.
 
 Notes:
-- SSH only; there is no API fallback transport. Windows remotes are out of scope.
+- SSH only; there is no API fallback transport. Windows desktops use the
+  installed OpenSSH client, but remote target accounts must still be POSIX;
+  Windows remote targets are out of scope.
 - Remote accounts need a POSIX shell and `ps`; Codex/Hermes reconciliation
   additionally needs `python3`, and their shell hooks need `curl`. Startup
   stays gated and reports a retrying state when a required dependency is absent.
--- a/packages/petdex-desktop-native/src/remote_ssh.zig
+++ b/packages/petdex-desktop-native/src/remote_ssh.zig
@@ -154,9 +154,10 @@ pub const Scratch = struct {
 /// children, but SIGTERM/crash paths cannot run it; without the wrapper,
 /// ssh is reparented to launchd and holds the remote port forever.
 pub fn installTunnelSupervisor(home: []const u8) bool {
-    // Windows launches OpenSSH directly. The remote health/lease command
-    // closes an orphan after the desktop disappears, so no local shell
-    // wrapper is installed or invoked.
+    // Windows launches OpenSSH directly. The remote command's `$PPID` is the
+    // sshd session, not the desktop process, so its health probe is what
+    // closes an orphan after three two-second misses (about six seconds).
+    // No local shell wrapper is installed or invoked.
     if (builtin.os.tag == .windows) return detect() != null;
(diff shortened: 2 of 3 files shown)


skb50bd:
@Railly, when you have a moment, could you please review this independent remote-SSH hardening slice? The macOS, Linux, and Windows checks passed, including Linux real-SSH writeback; the remaining Vercel status is authorization-only. Thank you.

Railly:
Reviewed the Zig. It holds up. `zig test src/remote_ssh.zig` passes 86/86 on this branch against 84 on base.

Checked the parts that looked risky and found them sound:

- `detect()` returning the `"ssh.exe"` sentinel on Windows is safe. The only two consumers, `main.zig:1761` and `remote_ssh.zig:160`, compare against null and discard the value, so no spawn path receives it.
- `Scratch` has no dangling slice. All six builders share one call site at `main.zig:1717`, in the same frame as the spawn at 1741.
- `writeArgv` with `resolve_link` disabled keeps `$target` well defined, and config relocation still works because non-executables keep the readlink branch. Reverting that line turns test 9 red, so the coverage is real.
- `windowsSshCandidate` handles `C:\`, UNC, quoted, trailing-slash and relative entries. The `> 3` guard is what stops `C:\` from degrading to `C:`.

Two notes.

The Windows comment on `installTunnelSupervisor` reaches the right conclusion for the wrong reason. `owner=$PPID` is the remote sshd session, not the desktop, so `owner_alive` does not fire when ssh.exe dies. What actually closes the orphan is `healthy()` failing its probe against the in-process hook server: 3 misses at 2s each, so exit 75 in about 6 seconds. Worth correcting the comment so the next reader does not rely on the stated mechanism.

`tunnelArgv` is the only runtime-dir builder without `chmod 700` (`remote_ssh.zig:294`). Not a defect today, since the state machine forces quiesce before tunnel and quiesce must exit 0. It is safe by call ordering rather than on its own, so it is worth hardening if that sequence changes.

I could not run the Windows binary or the docker SSH test here, so the orphan finding rests on mechanism reproduction with POSIX stand-ins plus reading the SDK teardown path, not an end-to-end Windows run.

skb50bd:
@Railly, thanks again for the careful review. I pushed 6d37faf addressing both notes: the Windows orphan-cleanup comment now describes the health-probe mechanism accurately, and the tunnel runtime directory is hardened with chmod 700 plus regression coverage. The Zig suite passes 86/86. When you have a chance, would you mind taking another look?

skb50bd:
@Railly this one got green signal from codex review. can we merge this?

skb50bd:
sorry, accidentally closed it.

Issue #717, opened 2026-08-17
[Submit fail] 若兰学姐


Milkkkk1204:
## ⚠️ BEFORE YOU SUBMIT THIS ISSUE

Without your pet files I cannot recover the upload. Please:

- [x] **Attach your zipped pet folder below** (drag-and-drop the .zip into the comment box). It must contain `pet.json` + `spritesheet.webp` (or `.png`).
- [x] If the pet has a backstory or tags I should add, paste them in a comment.

Issues without a zip get closed after 48h because there is nothing for me to import.

---

## What the form captured

- **Pet name:** 若兰学姐
- **Pet id:** limeacademia
- **Sprite size:** 1536×2288
- **Source:** folder
- **Error:** 上传失败：presign 404
- **User id:** `user_3I26JMjstBVTwTTQW3y5fGjnxJN`
- **Description:**
  > 戴黑框眼镜、半扎长发、学院西装和耳机的聪慧学院风桌宠。

Railly:
Found the cause, and it is our bug rather than your upload. Sorry for the trouble.

When a Clerk session is missing or expired, `POST /api/r2/presign` returns 404 instead of 401. The presign route itself is fine and never emits 404. The request never reaches it.

In `src/proxy.ts:99` we call `auth.protect()` before the `/api` early return on line 103, so protected API routes go through Clerk's page-oriented handling. Clerk redirects document requests to sign-in, but for a `fetch()` (`Sec-Fetch-Dest: empty`) it falls through to `notFound()`, which Next serves as a 404. The client reads that status and prints `presign 404`.

Verified against production just now:

```
POST /api/r2/presign        -> 404   (in isProtected)
POST /api/submit            -> 404   (in isProtected)
POST /api/pet-requests/image -> 401  (not in isProtected, behaves correctly)

POST /api/r2/presign, Sec-Fetch-Dest: empty    -> 404
POST /api/r2/presign, Sec-Fetch-Dest: document -> 307 to sign-in
```

Same URL and same second, so the only variable is one request header.

Your locale had nothing to do with it. The submit form calls the unprefixed `/api/r2/presign`, and `上传失败` only identifies the UI language. The form also still showed you as signed in because `useUser()` serves cached client state, which is why the report captured a real user id while the server saw no session.

[...]

Issue #716, opened 2026-08-17
[Submit fail] 若兰学姐


Milkkkk1204:
## ⚠️ BEFORE YOU SUBMIT THIS ISSUE

Without your pet files I cannot recover the upload. Please:

- [ ] **Attach your zipped pet folder below** (drag-and-drop the .zip into the comment box). It must contain `pet.json` + `spritesheet.webp` (or `.png`).
- [ ] If the pet has a backstory or tags I should add, paste them in a comment.

Issues without a zip get closed after 48h because there is nothing for me to import.

---

## What the form captured

- **Pet name:** 若兰学姐
- **Pet id:** limeacademia
- **Sprite size:** 1536×2288
- **Source:** folder
- **Error:** 上传失败：presign 404
- **User id:** `user_3I26JMjstBVTwTTQW3y5fGjnxJN`
- **Description:**
  > 戴黑框眼镜、半扎长发、学院西装和耳机的聪慧学院风桌宠。

Railly:
Duplicate of #717, filed 4 seconds apart from the same submit. Consolidating there so the reply stays in one place.

Issue #655, opened 2026-08-04
[Submit fail] 宇智波佐助


296277:
## ⚠️ BEFORE YOU SUBMIT THIS ISSUE

Without your pet files I cannot recover the upload. Please:

- [ ] **Attach your zipped pet folder below** (drag-and-drop the .zip into the comment box). It must contain `pet.json` + `spritesheet.webp` (or `.png`).
- [ ] If the pet has a backstory or tags I should add, paste them in a comment.

Issues without a zip get closed after 48h because there is nothing for me to import.

---

## What the form captured

- **Pet name:** 宇智波佐助
- **Pet id:** sasuke3
- **Sprite size:** 1536×2288
- **Source:** zip
- **Error:** 上传失败：Failed to fetch
- **User id:** `user_3HG8oMTPDXRDaCC65mmPLQbYKqx`
- **Description:**
  > 宇智波佐助的Q版宠物

henryjing96:
中国大陆地区上传可能需要开代理，建议您开代理重试下？

Railly:
Closing as resolved: `sasuke3` is now live on Petdex and its public sprite is available. Thanks for the report.

PR #721, opened 2026-08-20
fix: answer 401 instead of 404 on protected API routes

Railly:
## Problem

Signed-out or expired requests to `/api/r2/presign` get a 404, so the submit form reports `presign 404` and the upload dies. Reported in #717 (and #716, a duplicate of it).

`auth.protect()` ran at `src/proxy.ts:99`, before the `/api` early return on line 103. That sends protected API routes down Clerk's page-oriented path, which redirects a document request but calls `notFound()` for a `fetch()`. The presign route never runs, so its own 401 never gets a chance.

Verified against production:

```
POST /api/r2/presign         -> 404   (in isProtected)
POST /api/submit             -> 404   (in isProtected)
POST /api/pet-requests/image -> 401   (not in isProtected, correct)

POST /api/r2/presign, Sec-Fetch-Dest: empty    -> 404
POST /api/r2/presign, Sec-Fetch-Dest: document -> 307 to sign-in
```

Same URL and same second on the last pair, so one request header is the only variable. `/api/submit` and `/api/my-pets/*` carried the same defect.

## Change

Return for `/api` before `auth.protect()` runs. `/submit` is a page and keeps its redirect to sign-in.

This removes no protection. All seven routes the matcher named already call `auth()` and return 401 without a session, so they now emit that instead of a 404. The `/api` entries stay in `isProtected` as a backstop in case the ordering changes again.

## Not the cause

The reports came from `zh` users, so locale and geography both looked plausible. Neither holds. The submit form calls the unprefixed `/api/r2/presign`, and `上传失败` only identifies the UI language. Reproduced from Buenos Aires.

#655 is a different bug. Its error is `Failed to fetch` from the R2 PUT loop, not `presign 404`. The shared catch in `pet-submit-form.tsx:533` is what makes the two look alike.

## Test plan

- [x] `bun test src/proxy.test.ts`, 8 new tests
- [x] force-red: restoring the old order turns the ordering test red, and the fix turns it green
- [x] `bun run i18n:check`
- [x] Biome on changed files and `git diff --check`
- [x] `TELEMETRY_RATELIMIT_SECRET=mock-telemetry-secret bun --env-file=.env.mock run build`
- [x] `bun test --env-file=.env.mock`, 441 pass. The 2 CLI asset allowlist failures reproduce on a clean tree with the same command, so they are not from this branch.

## Follow-up, not in this PR

`pet-submit-form.tsx:473-481` prints the raw status. A 401 there should prompt re-auth instead.

### DIFF
--- a/src/proxy.test.ts
+++ b/src/proxy.test.ts
@@ -1,0 +1,54 @@
+import { describe, expect, it } from "bun:test";
+import { readFileSync } from "node:fs";
+
+// The proxy itself pulls in Clerk and next-intl, so we assert on its
+// source instead of importing it. What matters is an ordering the type
+// system cannot express: /api must return before auth.protect() runs.
+// Comments are stripped first, otherwise prose mentioning auth.protect()
+// would be read as the call site and the assertion would measure itself.
+const source = readFileSync(new URL("./proxy.ts", import.meta.url), "utf8")
+  .replace(/\/\*[\s\S]*?\*\//g, "")
+  .replace(/\/\/.*$/gm, "");
+
+describe("protected API routes bypass auth.protect", () => {
+  it("returns for /api before calling auth.protect", () => {
+    // Scope to the Clerk-backed middleware. The anonymous middleware
+    // above it has its own /api early return, and matching that one
+    // would pass no matter how the Clerk branch is ordered.
+    const start = source.indexOf("const clerkBackedMiddleware");
+    expect(start).toBeGreaterThan(-1);
+    const body = source.slice(start);
+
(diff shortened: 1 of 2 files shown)


Railly:
Verified on the preview deployment against production as the control.

```
preview (this branch)          production (main)
/api/r2/presign      -> 401    /api/r2/presign -> 404
/api/submit          -> 401    /api/submit     -> 404
/api/my-pets/approved -> 401
```

Same request in both cases: POST, matching Origin, no session. The three routes now reach their own handler and answer 401.

Issue #703, opened 2026-08-14
Windows x64 v0.8.0: pet renders but ignores all mouse input


milesc1616-hue:
**Environment**
- OS: Windows 11 (screen 1707x1067, likely VM/remote desktop)
- App: petdex-desktop-win32-x64.exe v0.8.0 (SHA256 419243A8...)
- Pet: Stamps (installed via `npx petdex install stamps`)

**What happens**
1. The pet sprite renders correctly (orange/white tabby visible).
2. Left-click: no pat reaction.
3. Right-click: no context menu (Open Settings / Close Pet).
4. Dragging: window does not move.

**What I verified**
- spritesheet.webp decodes fine via WIC (Microsoft Webp Decoder, 1536x2288).
- `~/.petdex/desktop-native-settings.json` has `active_pet: "stamps"`.
- Process stays alive; the window exists, is visible, enabled, layered,
  not WS_EX_TRANSPARENT, not WS_EX_NOACTIVATE.

**Expected**: left-click should pat the pet (#617); right-click should
show the context menu; drag should move the window.

Railly:
Traced this. Your diagnosis was right to rule out the window flags: the window does receive the input, but nothing on Windows is wired to consume it. The cause is in the Native SDK, not in this repo.

Every pet interaction on Windows goes through one call. `packages/petdex-desktop-native/src/main.zig:2496`:

```zig
const read = fx.moveWindow("main", 0, 0, false) orelse return;
```

Hover, the drag branch, the press detection, and the pat reaction all sit downstream of that line. `moveWindow` returns null when the platform has no `move_window_fn`, and at the pinned SDK commit `c0b10d0` the Windows platform never registers one:

```
src/platform/windows/root.zig   move_window_fn: 0 occurrences
src/platform/macos/root.zig     move_window_fn: 1 (line 592)
```

So the `orelse return` fires on every poll and left-click and drag are both dead. The comment at `src/main.zig:3755` says Windows and macOS share the app-owned drag path, and that is the origin of the bug: the path depends on a capability only macOS has. Linux avoids it by taking a separate branch at `src/main.zig:3749`.

Right-click has its own cause. `src/platform/windows/root.zig:371` reports `context_menus => false` where macOS reports `true` at `macos/root.zig:678`, so `.context_menu = &pet_menu` on the pet root is a no-op on Windows by design. TrackPopupMenu is named there as the intended future seam.

[...]

PR #723, opened 2026-08-20
ci: click the pet on Linux and Windows and assert it reacts

Railly:
## Why

The Linux and Windows legs already launch the app, photograph it, and drive a hook round trip. What they never do is send the pet a mouse event. The workflow header says so directly: on Windows, "topmost ordering, the black color key against real wallpaper, and click-through are compile-and-launch verified only".

#703 is that gap. On Windows the pet renders correctly and ignores every click, drag and right-click, and the Windows leg stayed green through all of it. The cause is in the Native SDK: `windows/root.zig` registers no `move_window_fn` while `macos/root.zig:592` does, and `src/main.zig:2496` routes every pet interaction through `fx.moveWindow(...) orelse return`.

A green build proving the binary starts says nothing about whether input reaches it. This closes that.

## What it does

Downloads cuse on the Linux and Windows runners, taps the pet twice, and asserts the frame changed both times. cuse drives real OS input, XTEST on X11 and SendInput on Win32, rather than synthesizing events inside the app.

Two taps rather than one on purpose. A pat alternates the pet between jumping and waving, so a single changed frame could be an idle animation landing elsewhere, but two consecutive changes cannot.

The Linux leg gets its own Xvfb, since the hook-server step starts one and exits. On Windows the binary is saved as `cuse.exe`, because Git Bash will not execute an extensionless file.

## What fails, and how it reads

If input stops reaching the pet, the step prints the verdict for each tap and the app log, then fails with a message naming #703. If the window cannot be matched at all, cuse errors with the list of visible windows, which is the diagnostic you want at that moment.

## Verified before pushing

- `bash -n` on the extracted step, and the YAML parses
- cuse `v0.1.0` exists with `cuse-linux-x64` and `cuse-windows-x64.exe` published, along with `SHA256SUMS`
- the checksum extraction was run locally against the real binary and reports `cuse: OK`. My first version used `sed` and produced a three-space separator that failed to verify, so it is `awk` now
- `diff --json` returns `data.verdict` as `SAME` or `CHANGED`, which is the field the step reads rather than grepping the line
- the window title is `Petdex`, from `src/main.zig:59`
- `click --window` fails loudly with the visible-window list when nothing matches

## What this does not prove

It asserts that a click produces a visible change, not that the change is the correct sprite. It also does not cover drag or right-click, which have their own causes on Windows and stay open under #703.

I could not execute this against a real Windows or Linux desktop from my machine, so the first run on CI is its real test. Worth watching that run before merging.

### DIFF
--- a/.github/workflows/desktop-native-ci.yml
+++ b/.github/workflows/desktop-native-ci.yml
@@ -831,11 +831,21 @@ jobs:
             linux)
               asset=cuse-linux-x64
               bin=./cuse
-              # The hook-server step starts its own Xvfb and exits, so
-              # this one needs a display of its own.
+              # cuse reads the screen through xwd, which is not on the
+              # runner by default.
+              sudo apt-get install -y x11-apps xvfb
               export DISPLAY=:99
-              Xvfb :99 -screen 0 1280x800x24 &
-              sleep 3
+              # :99 may still be up from the hook-server step. Starting a
+              # second server on a live display is a fatal error, so only
+              # start one when nothing answers.
+              if ! xdpyinfo -display :99 >/dev/null 2>&1; then
+                Xvfb :99 -screen 0 1280x800x24 &
+                for _ in $(seq 1 30); do
+                  xdpyinfo -display :99 >/dev/null 2>&1 && break
+                  sleep 1
+                done
+              fi
+              xdpyinfo -display :99 >/dev/null 2>&1 || { echo "no display on :99"; exit 1; }
               ;;
             windows)
               asset=cuse-windows-x64.exe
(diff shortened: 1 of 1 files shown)


Railly:
Ran this four times on real runners. Summary of what it found, since the outcome changed what the PR is worth.

What now works: cuse downloads and verifies against the published SHA256SUMS on both platforms, runs natively (`"os":"windows"`, `"os":"linux"`), and the app starts under the step with a fresh token.

Four real bugs surfaced along the way, three of them mine:

1. Linux needs `x11-apps`, since cuse reads the screen through `xwd`.
2. Starting a second Xvfb on the live `:99` is a fatal error, so the step now reuses the display when one answers.
3. On Windows, git-bash's `~` is not `$USERPROFILE`, the same trap the "Install a pet (Windows)" step already documents. My readiness gate was watching the wrong path, and an old token from an earlier step made it return instantly. It clears the token first and exports `HOME=$USERPROFILE` now.
4. The window title is not what I assumed: GTK reports `petdex-desktop-native`, not the `Petdex` from `src/main.zig:59`. Matched per platform now.

The remaining blocker is not in this repo. On a bare X11 display, `cuse windows` returns `{"ok":true,"detail":"no titled windows","data":[]}` while `xwininfo` in the same job lists `0x200008 "petdex-desktop-native" 253x115+0+97`. With no window to aim at, the click cannot land. Filed as crafter-agents/cuse#65, with the likely cause being that a bare Xvfb has no window manager and so no `_NET_CLIENT_LIST` to enumerate.

Windows has a separate blocker of its own: `3 pet(s) installed but none decoded; the sheet may be corrupt`, so no window opens there either. That one may be worth its own issue.

So the step is `continue-on-error: true` for now and the three legs are green. It runs and reports on every push without gating. The comment above it says to flip that off once Linux can enumerate windows, which is the point where this actually starts catching #703-class regressions.

Merging it in this state is still worth it: the scaffolding is proven on real runners, and the failure it prints is now an accurate diagnosis rather than a mystery. But it does not gate anything yet, so it is fair to hold it until cuse#65 lands.

PR #781, opened 2026-09-14
fix(cli): defer submit entry until module evaluation completes

### LINKED ISSUE #778: petdex submit crashes on CLI 1.3.0: entry point runs before module evaluation finishes (license list is undefined)
## Summary

`petdex submit` is broken in CLI **1.3.0**: it crashes before any network call, so no pet can be submitted through the CLI. Both the interactive and the `--license` paths fail.

```console
$ npx petdex@latest submit ./gabiru
└  petdex: Cannot read properties of undefined (reading 'map')

$ npx petdex@latest submit ./gabiru --license cc-by
└  petdex: Cannot read properties of undefined (reading 'find')
```

## Root cause

In `dist/petdex.js`, the entry point is invoked **while the module is still evaluating**:

```js
// ~line 77 of the bundle
Q_().catch((e) => { q(`petdex: ${e.message}`), process.exit(1) });

// ~line 84, further down in the same module
var fa = [
  { value: "cc0",  label: "CC0 — public domain, no credit needed" },
  { value: "cc-by", label: "CC BY — any use, with credit" },
  // ...
];
```

`Q_()` runs synchronously up to its first `await`. For `submit` that path reaches `ob()`, which touches the license list **before** any `await`:

```js
let i = ab(e), n;
if (i !== null) { let g = fa.find((y) => y.value === i); /* ... */ }
else { let g = await sn({ ..., options: fa.map(...) }); }
```

At that moment the module has not yet reached the `var fa = [...]` statement. The `var` binding is hoisted but still `undefined`, hence `.find`/`.map` of undefined. With `--license` it hits `.find`; without it, `.map`.

Any command whose synchronous prologue reads a `var` declared after the entry call is exposed to the same ordering hazard — `submit` is just the one that does it first.

## Reproduction

1. `npx petdex@latest login`
2. Create a folder with a valid `pet.json` and a 1536x1872 `spritesheet.webp`
3. `npx petdex@latest submit ./that-folder`

Crashes for me on every attempt, with and without `--license`.

## Workaround (verified)

Defer the entry call by one microtask so module evaluation completes first. One-line change to the installed bundle:

```js
queueMicrotask(() => {
  Q_().catch((e) => { q(`petdex: ${e.message}`), process.exit(1) });
});
```

With that patch the submission went through normally and the pet reached the review queue.

## Suggested fix

Move the entry invocation to the end of the module, or wrap it as above, so no command can observe a partially evaluated module. Hoisting the license list to the top of the file would fix `submit` specifically, but the ordering hazard would remain for other commands.

## Environment

- petdex CLI 1.3.0 (latest on npm)
- Node.js v22.22.3, npm 12.0.2
- Ubuntu, Linux 7.0.0-31-generic

ege-arhan:
Problem: `petdex submit` crashes on CLI 1.3.0 before any network call — `Cannot read properties of undefined (reading 'map'/'find')`. The bundled entry `main()` can run while the module is still evaluating, so `LICENSE_CHOICES` is still in TDZ/undefined when `cmdSubmit` reads it synchronously (.find with --license, .map without).

Solution: wrap the entry invocation in queueMicrotask so module evaluation always completes first, even if a future bundler reorders the bundle: `queueMicrotask(() => main().catch(...))`. Source fix in packages/petdex-cli/bin/petdex.ts (dist not hand-patched).

Impact: restores `petdex submit` (interactive + --license paths) for this fast-growing repo (4096 stars, ~600 in May 2026).

Evidence: esbuild bundle verified — queueMicrotask wrap present at bundle tail; smoke test `submit ./some-pet --license bogus` exits 1 with 'Unknown --license bogus. Options: cc0, cc-by, cc-by-sa, cc-by-nc, all-rights-reserved' (no undefined crash); tsc --noEmit clean; git diff --check clean.

Fixes #778


ege-arhan:
Vercel deployment requires authorization from maintainer. Code change ready, waiting for deploy approval.

